Issue #261, opened 2014-07-13
More modals needed for confirmation


Martii:
Currently when sending something to the graveyard we aren't giving an option to cancel/close.  Also setting the "Reason" why... it would be helpful to have a modal popup, similar to Delete Script, to enter in the reason.

We do currently have allocated in the DB a space for this but we aren't utilizing it.

Flagging could also have a reason stored but I don't currently see that capability.

---

##### Progress

* [x] Basic modals on User and Script removal
* [x] Reason in modals *(See also https://github.com/OpenUserJs/OpenUserJS.org/issues/261#issuecomment-48852243)*
* [x] Default reasons datalist *(See https://github.com/OpenUserJs/OpenUserJS.org/issues/261#issuecomment-69467296 and useful to narrow results for #490)*

Martii:
There are currently a few possibilities to get this particular issue closed:

1. Do nothing with the data list and let a user.js handle this on a per Moderator and up basis.
2. Add the data list into the `scriptModals.hmtl` and `userModals.html` ... this could help prospective Moderators and up to standardize on nomenclature, less typing, and easily searchable/sortable but also increases that particular modal view.
3. Add the data list into a json and read in dyanmically... this presents the issue of file naming and where to put the .json file. If this is done please mention the file name structure and where. This could also help prospective Moderators and up to standardize on nomenclature, less typing, and easily searchable/sortable.
4. Something else?

Opening up to needs discussion to put in others 2 cents. Please note if I don't get a response I may end up deciding on what to do here. e.g. this is a pseudo poll. Thanks.

sizzlemctwizzle:
I like 3 the best, but 2 seems easier to get up and running. My suggestion is to start with 2 and eventually use 3.

Martii:
Agreed on 3... I finished 2 a little while ago and refinished 3 in separate commits. With that process that helped me figure out where the "optimum" placement of the .json files should go I think plus an epiphany on it's structure/inclusion too. :) Thanks @sizzlemctwizzle 

Publishing PR shortly.

Martii:
Heh... GH currently has the commit references backwards. :\

Martii:
;) Now they've fixered the reference order of commits... love being watched! Good thing I saved a snapshot of when it was backwards. ;)

PR #513, opened 2015-01-10
Support existing incompleted "reason" value for remove user and script

Martii:
* One reserved phrase of "User removed" *(any casing)* for automatic content removal when a User is removed.
* Basic error checking... Moderators and above should be concise.
* Abandoning previous UI since portable devices would barely show the input box... plus a few things have changed since initial attempt.
* Leaving out the data list for now until it can be decided where to physically put and name it *(them?)* in the project tree... perhaps time for a user.js.

Applies to #261

### DIFF
--- a/controllers/remove.js
+++ b/controllers/remove.js
@@ -11,18 +11,40 @@ var Script = require('../models/script').Script;
 var User = require('../models/user').User;
 var destroySessions = require('../libs/modifySessions').destroy;
 
+var formidable = require('formidable');
+var statusCodePage = require('../libs/templateHelpers').statusCodePage;
+
 // Simple controller to remove content and save it in the graveyard
 exports.rm = function (aReq, aRes, aNext) {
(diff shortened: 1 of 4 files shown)


controllers/remove.js
```diff
@@ -11,18 +11,40 @@ var Script = require('../models/script').Script;
...
 
+  var form = null;
+  var reason = null;
+
+  if (!/multipart\/form-data/.test(aReq.headers['content-type'])) {
+    return aNext();
+  }
+
+  form = new formidable.IncomingForm();
+  form.parse(aReq, function (aErr, aFields) {
```
jerone: formidable.`parse()` is [async](https://github.com/felixge/node-formidable/blob/master/lib/incoming_form.js#L49).

Martii: Good to know... will get that after slumber too. Have another use case to get too. Thanks.

PR #519, opened 2015-01-11
Issue 261 data list

Martii:
See commits for detailed specifics

Root issue is #261

### DIFF
--- a/controllers/remove.js
+++ b/controllers/remove.js
@@ -74,7 +74,7 @@ exports.rm = function (aReq, aRes, aNext) {
                 return aNext();
               }
 
-              // Destory all the sessions belonging to the removed user
+              // Destroy all the sessions belonging to the removed user
               destroySessions(aReq, aUser, function () {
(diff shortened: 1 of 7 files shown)


Martii:
Going to leave this available for comment until Sunday evening'ish in case someone wants different default reasons added or a bug detected... if nothing replied then will merge it.

Martii:
I wanted to explain "Undocumented Auto-Like" here.

These are scripts that Auto-Like something **without the Users permission**. Currently we have at least one script now that actual seems nice about Auto-Liking and documents the Code itself and doesn't just add their own personal group ids and user ids on FB to improve their stats. **This is okay in my book* because it is clear and concise. However those scripts should always be watched in case they get overzealous and don't document in plain english what groups and users they are auto-liking.

Martii:
There's now an example in the Graveyard with what was detected. This particular author and script also had an undocumented auto-like since it claims to increase likes... e.g. not documented as doing this.

PR #529, opened 2015-01-13
Fix Library require links to be protocol specific

Martii:
* Currently they show as http which our default is https ... so if the request comes from a https show https, etc.

**NOTE**: Btw this is cheating using the host header to get the port inclusion instead of crafting it from the components.


Martii:
Hmmm now that's peculiar... it still says http in libs :\ May have to make this a static https here.

Issue #531, opened 2015-01-13
Mirror paginations at top


Martii:
So it's about time to reconsider and revisit this. This should be relatively simple to duplicate but I wanted to read what others had to say about this.

@BastianKanaan might want to pipe in here since that script has the [largest discussion](https://openuserjs.org/scripts/BastianKanaan/MyFreeFarm_Berater/issues/General_discussion) to date that I've seen on OUJS.

jerone:
From a design/layout perspective I'm :-1: on this; [1](http://ux.stackexchange.com/questions/16530/whats-the-design-rationale-behind-putting-the-pagination-of-the-bottom-end-of-th), [2](http://ux.stackexchange.com/questions/19269/pagination-placement-of-navigation-control-relative-to-search-results?lq=1), [3](http://ux.stackexchange.com/questions/31603/paging-buttons-position-top-bottom-or-both)

Martii:
Try that link on a portable device with the URL I posted above and might change your mind. Nobody in their right mind would want to swipe that many times on a touch screen... ever heard of carpal tunnel?! ;) 

I'm still tweaking which is why this is still open. Yes there will be some more white space to dark space in that area... the advantage of being able to go to the end supersedes visual eye-candy in this case.

Sizzle already gave the thumbs up and I agree with it. I'll do my best to minimize the gaping gaps that have been recently introduced *(started in #477 and beyond)* but it will take some time to iron out.

Thanks for the links... [3](http://ux.stackexchange.com/questions/31603/paging-buttons-position-top-bottom-or-both) seems to be the most readable and logical so far.

jerone:
@Martii commented on [14 jan. 2015 09:13 CET](https://github.com/OpenUserJs/OpenUserJS.org/issues/531#issuecomment-69882539 "2015-01-14T08:13:03Z - Replied by Github Comment Enhancer"):
> Try that link on a portable device with the URL I posted above and might change your mind. Nobody in their right mind would want to swipe that many times on a touch screen... ever heard of carpal tunnel?! ;) 

Valid use case; then use Bootstrap for [only showing on small devices](http://getbootstrap.com/css/#responsive-utilities-classes).

Martii:
And ones scroll finger on their desktop, which isn't a small device, invalidates that completely. All reputable forum softwares have pagination to the end immediately. The concept of making one scroll to the end of a page just to get to pagination is, digressing for a moment, quite frankly moronic. :)

[...]

PR #537, opened 2015-01-14
Fix the bottom existing pagination at `/forum`

Martii:
* Panel was initialized with pagination that skewed off center for additions and compounded by #443. This should align the existing pagination properly.

Applies to #531


Martii:
Btw there are more of these missing the .panel-group but since this particular page has other bootstrap elements on the **left** it definitely needed them.

Issue #249, opened 2014-07-09
Establish a build/exec/deploy system


Martii:
~~Fork a command line minifier for dev tools on the OpenUserJS account~~

~~This will be needed in a while to minify what we want to before deploying along with a `deploy.sh` that is kept with sizzle and maybe here... depends on the secrets that need to be kept.~~

Dev and GH source should remain unminified at all times... only time minified should be existing is on deployment/testing... this may require some extra time with making sure minification doesn't clobber anything... usually they are quite good at what they do by now but to be safe an extra testing measure should be done.

This is probably a long term issue since it's going to take some time to undo what's been done.

---

We could also use, instead,  a shell script with git directly to a folder and then use some .sh scripts to move things to the appropriate locations... keeping in mind .gitignore . e.g. we really need a build.sh system and exec.sh type system.

---

This section may change constantly...
- [x] VPS
- [x] Initialization system... perpetual tweaking possibly. Invoked via `$ npm install` or `$ node dev/init.js`
  - [ ] Look into https://github.com/alexanderGugel/ied as alternative
- [x] Exec... Invoked via `$ npm start` or `$ node app.js` ... process management roughed in.
- [x] Deploy... `$ git pull` on VPS
- [ ] Minify CSS on the fly... partial... needs ~~enabling of cache system and placement on the VPS tree~~ with automated cleanup
- [ ] Minify JavaScript on the fly... partial... ~~needs enabling of cache system and placement on the VPS tree~~ with automated cleanup... wants #432
- [x] LESS caching _(maybe... probable that express-minify is already doing this. Also *bootstrap*@4.x will use SASS)_
- [ ] Minify HTML
- [ ] See if [*nock*](https://www.npmjs.com/package/nock) is viable for testing requests that aren't normally available in dev
- [ ] ...

Martii:
@cletusc Cc: @sizzlemctwizzle and anyone else too

Does anyone have a preference on the file cache for the release version of *express-minify*? This affects dev as well as pro.

Possible locations are:
1. `./.cache/express-minify` *(static or dynamically created?)*
2. `SystemTempSpecialFolder/openuserjs.org/cache/express-minify` *(dynamically created... optional `.oujs` short name instead of full?)*
3. `HomeDirectorySpecialFolder/.cache/openuserjs.org//express-minify` *(dynamically created... optional `.oujs` short name instead of full?)*
4. Other physical disk location... *(please specify)*

[...]

PR #540, opened 2015-01-15
Update *bootstrap* dep to latest

### LINKED ISSUE #379: Why are we not using bootstrap > 3.1.1?
Just for grins I updated [bootstrap](https://github.com/twbs/bootstrap) dev locally to 3.2.0 and it appears in a quick scan to be okay to use e.g. my normal testings it appears to be okay... not fully tested though.

This question is primarily directed towards @Zren

Martii:
* `z-index` was the culprit for previously mentioned modal failure
* Watchpoint for *bootstrap* 3.1.1 changes to 3.3.1 ... if necessary will swap out entire modal sections of CSS with upstream as they do work and are more compact but reserving that task for #249

Closes #379

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "async": "0.9.0",
     "aws-sdk": "2.0.21",
     "body-parser": "1.10.1",
-    "bootstrap": "3.1.1",
+    "bootstrap": "3.3.1",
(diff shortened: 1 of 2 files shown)


Martii:
Rolling back to 0.1.6-37. Comment popup wants to self-hide. Added needs mitigation to #379.

PR #544, opened 2015-01-20
Compile, parse, and cache *bootstrap* 3.1.1

Martii:
* Use *less-middleware* for compiling now until *express-minify* can be retested and possibly more RFEs. Methodology of *express-minify* has been tested to be cleaner with caching... but there's a new update and will have to run all those tests again. Project tree structure should be compatible with either one.

Applies to #249 and should be near parallel change for #379 

---

@Zren *(and anyone else too)*
I would appreciate it if you would pick over the .less files and make sure nothing is missing. I've been testing it for about an hour and I can't see any visible difference *(yet)*. I've added my changes to `public/less/bootstrap/oujs-bootswatch.less` and put in the short commit hash for now. This can use some improvement down the line but this is needed to move forward with the issues referenced... hence not closing yet. Thanks for your assistance.


Zren:
Does the compiled CSS file work from a fresh clone? As its in the ignore file.

I'll test it out later tonight.

Martii:
> As its in the ignore file.

It's in the ignore because that's how it caches... it creates the .css version of the .less file.

> Does the compiled CSS file work from a fresh clone?

We have write access on nodejitsu but it's also a unit test here to see what happens between snapshots. This hasn't been deployed yet because I wanted to make sure that all CSS classes were included. I don't care right now about the variable substitutions of what I added... we can all work on that later... but checking if any classes are missing is important before deploy and snapshot switch testing. *(the latter may take some time)*.

Martii:
Reply soon or forever hold your peace. :) Merging soon.

PR #564, opened 2015-02-05
Silently migrate from OpenID to OAuth2 authentication for Google.

sizzlemctwizzle:
Closes #484.

PR #565, opened 2015-02-05
Bump project version

Martii:
See also:
* https://github.com/OpenUserJs/OpenUserJS.org/pull/564#issuecomment-73019449

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "OpenUserJS.org",
   "description": "An open source user scripts repo built using Node.js",
-  "version": "0.1.6+0000000",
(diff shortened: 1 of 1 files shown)


Martii:
Nodejitsu isn't paying any attention to this again on snapshot creation... e.g. version on pro showing will probably be 0.1.6-51 instead of the correct 0.1.7 ... this might self correct at some point but keeping dev at this.

Martii:
Attempting `-r 0.1.7` on deploy to force the version switch. Repeated failed attempts *(-52 and -53)* to have nodejitsu auto read in existing semver `version` name... this might work... will update this comments status.

PR #602, opened 2015-03-18
Redirect the user to their previous page after login/logout.

sizzlemctwizzle:
Works for me locally. I'm not sure if it will work in production. It uses the "Referer" header and I'm not sure if Nodejitsu handles that properly.

### DIFF
--- a/controllers/index.js
+++ b/controllers/index.js
@@ -8,6 +8,7 @@ var isDbg = require('../libs/debug').isDbg;
 //
 var async = require('async');
 var _ = require('underscore');
+var url = require('url');
 
 var Discussion = require('../models/discussion').Discussion;
 var Group = require('../models/group').Group;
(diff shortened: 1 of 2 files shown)


sizzlemctwizzle:
Only one way to know. Deploying...

sizzlemctwizzle:
Tested and it apparently works!

controllers/index.js
```diff
@@ -151,13 +152,30 @@ exports.home = function (aReq, aRes) {
...
 
+// Get the referer url for redirect after login/logout
+function getRedirect(aReq) {
+  var referer = aReq.get('Referer');
+  var redirect = '/';
+
+  if (referer) {
+    referer = url.parse(referer);
+    if (referer.hostname === aReq.hostname) {
+      redirect = referer.path;
```
Martii: Was there a reason why you chose to omit search query and hash e.g. `?yadda=ying#comment-someHash`?

Martii:
Nice addition. :)

Martii:
Btw testing if referer is blocked momentarily to see what happens.

Martii:
Blocked `referer` header redirects to `/` ... so perfect.

Issue #604, opened 2015-03-19
Don't create sessions for update checks, update installs, and library downloads


sizzlemctwizzle:
Right now our production database is relatively small, with the exception of session storage. Express Mongo creates a session for anyone who accesses OUJS. This is problem for us because we are pounded with requests by GM and TM for update checks, update installs, and library downloads, which means a session is created for every instance of GM or TM that has a OUJS script installed. It doesn't matter if these users never login, register, or even visit the site, they all have sessions created for them. The solution to this problem is to find a way to bypass the session logic for `/src/` paths so that session are not created.

This issue is blocking because there were 900,000+ session records when I cleared out the session data today, and within half an hour there were already +2,000.  We need to keep session storage as small as possible for clear scale-ability issues. No new non-bugfix PRs will be merged until this issue is resolved. The site can't move forward if the database is exploding.

Martii:
Optimization wise #285 will affect this as well with inclusion of *pegjs* as the object will become much larger... so that is a consideration too. This can be a reason why I mentioned at one point we should store the raw metadata and serialize it to an object on the fly... although I haven't done any real comparisons just yet.

sizzlemctwizzle:
I'm working on a patch right now that should help until we can find something better.

sizzlemctwizzle:
That PR should work, but it definitely isn't the smoothest way to solve the problem.

sizzlemctwizzle:
Well my patch seems to have slowed things a bit. It would be nice if we could only create session documents for users when/if they log in. Creating a document for every browser that accesses the site seems crazy. It also seems like a vulnerability.

PR #605, opened 2015-03-20
Temporary patch for #604.

sizzlemctwizzle:
Don't create session records for source and meta requests by intercepting them before the session middleware is used. #604


sizzlemctwizzle:
Confirmed that it works. No cookies are saved when accessing the paths it intercepts.

PR #611, opened 2015-03-31
FB passport update

Martii:
* *passport-facebook* - **UNTESTED LOGIN BUT DEPLOYED** and awaiting final test from @sizzlemctwizzle

**NOTE**:
* *passport-google-oauth* requires Google+ API added to OUJS account and can not be updated even without migration of scope. @sizzlemctwizzle will need to do this.


Martii:
Passed test login from sizzle.

Issue #486, opened 2014-12-06
Migrate the scripts in S3 to use `script._id` instead of the path value


Martii:
Currently we are unable to upgrade the _aws-sdk_ dependency because "something" changed that makes [`getObject`](http://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#getObject-property) to fail.

This has been explained in detail at aws/aws-sdk-js#430 . With all the test deploys done trying to figure out what the steps to reproduce are it appears that normalizing the casing when we save to the DB is an option.

Looking for more options related to advancing the goal of updating the dep... e.g. perhaps we could validate the current `aReq.params.username` against the real username _(name)_ that is typically in the Raw JSON Data for an account. e.g. a lookup to correct the casing [around here](/OpenUserJs/OpenUserJS.org/blob/e0a03458f957cb199227e2ce033a6d14103cc2a1/controllers/scriptStorage.js#L47-L61)... but I suspect this is going to need to be a DB migration with a BLOCKING label until it is finished.

See also:
- https://github.com/aws/aws-sdk-js/issues/397#issuecomment-60807108 error trap suggestion for loosely related
- https://github.com/OpenUserJs/OpenUserJS.org/blob/71ef808762ccf14a048c8a639d6d620e4e4e91e4/controllers/script.js#L350 Probably related and should be fixed

Test points:
- https://openuserjs.org/scripts/Marti/oujs_-_Meta_View/source vs https://openuserjs.org/scripts/marti/oujs_-_Meta_View/source e.g. manual url entry in the addy bar
- https://openuserjs.org/scripts/JRI/Geocaching_Map_Enhancements/source Since `jri` is currently in the DB not `JRI`
- https://openuserjs.org/users/jri  _([http://localhost:8080/users/jri](http://localhost:8080/users/jri) with local pro)_

Martii:
Hmmm... retested this randomly with [aws-sdk-js@2.1.25](https://github.com/aws/aws-sdk-js/releases/tag/v2.1.25) and doesn't appear to crash dev nor local pro... not sure on actual production.  All prior test points appear to be okay. :\

@sizzlemctwizzle 
Any objections to trying it on production? I'll give this a day or two for any responses.

sizzlemctwizzle:
[...]

PR #626, opened 2015-04-27
Try for *aws-sdk-js* (*aws-sdk*) update

Martii:
Appears to work in dev and local pro... not sure about production

See also:
* https://github.com/OpenUserJs/OpenUserJS.org/issues/486#issuecomment-96524838


Martii:
This PASSed on testing.. however we did have a *mongoose* error restart... will put on watch list.

PR #635, opened 2015-05-28
Nodejitsu shutting down. Modifications for new host.

sizzlemctwizzle:
Well the Joyent team that maintains Nodejitsu got purchased by Goddady to develop their website builder. The result is that Nodejitsu is shutting down in August. I've moved the site to a new host and made the necessary modifications to the code. The site is likely more portable as a result. [openuserjs.org](https://openuserjs.org/) is currently running on this branch.

### DIFF
--- a/app.js
+++ b/app.js
@@ -46,6 +46,7 @@ var server = http.createServer(app);
 var secureServer = null;
 
 app.set('port', process.env.PORT || 8080);
+app.set('securePort', process.env.SECURE_PORT || null);
 
 // Connect to the database
 mongoose.connect(connectStr, dbOptions);
@@ -55,7 +56,7 @@ db.once('open', function () {});
 var sessionStore = new MongoStore({ mongooseConnection: db });
 
 // Force HTTPS
-if (app.get('port') === 443) {
+if (app.get('securePort')) {
   sslOptions = {
     key: fs.readFileSync('./keys/private.key'),
     cert: fs.readFileSync('./keys/cert.crt'),
@@ -74,8 +75,8 @@ if (app.get('port') === 443) {
     aNext();
   });
 
-  server.listen(80);
-  secureServer.listen(443);
+  server.listen(app.get('port'));
+  secureServer.listen(app.get('securePort'));
 } else {
   server.listen(app.get('port'));
 }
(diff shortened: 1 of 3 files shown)


Martii:
I suggest giving this a couple of days running on production before merge to keep the transition of dev source here clean... I'll run some more tests in dev/local pro to ensure everything appears to be working with my available authentication mechanisms and tests. I'll halt dep update testings until this is done. e.g. this should be blocking for a few days.

sizzlemctwizzle:
I plan on canceling my nodejitsu subscription on Sunday so I don't get
billed for June. We can wait until then to merge.

app.js
```diff
@@ -38,29 +38,47 @@ var sessionSecret = process.env.SESSION_SECRET || settings.secret;
...
-db.once('open', function () {
-  app.listen(app.get('port'));
-});
+db.once('open', function () {});
 
 var sessionStore = new MongoStore({ mongooseConnection: db });
 
 // Force HTTPS
-if (app.get('port') === 443) {
+if (app.get('securePort')) {
```
Martii: @sizzlemctwizzle 
So you are doing this programmatically now from the environment now instead of what was done at https://github.com/OpenUserJs/OpenUserJS.org/commit/932d1c906d70d6b305231d1ad381f60a7a717ea3 ? e.g. which no longer allows http://openuserjs.org ?

Martii:
Had to reset file perms on dynamicly generated CSS at [.../public/less/bootstrap](https://github.com/OpenUserJs/OpenUserJS.org/tree/8f231154f5a4c2ecf4aa3a6d21a4aba10e30eb80/public/less/bootstrap)/oujs.css

We may need to watch that.

Martii:
Reset timezone to UCT... better for most log rotation around daylight savings time.

app.js
sizzlemctwizzle: If you don't set the SECURE_PORT environment variable then it runs using
http on PORT.

Martii:
+1

Martii:
> note not all logs are showing this change and may be configured/initialized elsewhere... might need a restart of the server instead of a restart of the project.

A server reboot fixed a log clock however reboot also killed the web service... messaged @sizzlemctwizzle to restart ... apologies for the downtime.

sizzlemctwizzle:
Back up again.

Issue #285, opened 2014-07-29
Decide the fate of custom site/project prefixing in the UserScript metadata block


Martii:
In relation to greasemonkey/greasemonkey#1963 I see 3 possibilities so far over the past few days to address an impending conflict with everyones project.

``` javascript
// ==UserScript==
// @run-at             document-end
// @namespace          http://localhost.localdomain
// @name               RFC 2606§3 - @license and @licence (recovered) Unit Test
// @name:de-DE         RFC 2606§3 - @license und @licence (wiederhergestellt) Komponententest
// @name:fr            RFC 2606§3 - @license et @licence (récupéré) Test unitaire
// @description        Tests out non-unique keys
// @description:de-DE  Tests, nicht eindeutige Schlüssel
// @description:fr     Tests de clefs non unique
// @copyright          2007+, Marti Martz (http://userscripts.org/users/37004)
// @license            GPL version 3 or any later version; http://www.gnu.org/copyleft/gpl.html
// @version            0.0.0
// @include            http://www.example.com/*
// ==/UserScript==

// ==OpenUserJS==
// @author Marti
// @collaborator jerone
// @collaborator sizzle
// @collaborator Zren
// @hash 131a5eddc53f4018d6b5ee76d8543cb3cb4e50fa
// @somefutureidea somevalue
// ==/OpenUserJS==
```
* Has the disadvantage that any user.js engine may not read it. **EDIT**: ...but at the same time resolves any external conflicts and is in the spirit of optional collaboration instead of mutual collaboration.
* OUJS could store the **raw formatted metadata blocks text** instead of the single metadata block Object... and a migration routine put in place or applied. **EDIT**: We should do at least the latter part of this anyways for greasemonkey/greasemonkey#1963 support no matter what site/project prefixing we support. I highly recommend the former as well but depends on how we finalize the stored Object or not.

or

``` javascript
// ==UserScript==
// @run-at             document-end
// @namespace          http://localhost.localdomain
// @name               RFC 2606§3 - @license and @licence (recovered) Unit Test
// @name:de-DE         RFC 2606§3 - @license und @licence (wiederhergestellt) Komponententest
// @name:fr            RFC 2606§3 - @license et @licence (récupéré) Test unitaire
// @description        Tests out non-unique keys
// @description:de-DE  Tests, nicht eindeutige Schlüssel
// @description:fr     Tests de clefs non unique
[...]

PR #670, opened 2015-07-20
Remove generated `null` name generation

Martii:
* Distinctly create unique objects in this portion code to maintain the object order for the POC
* Trying start points... documentation appears a bit vague... will revert this portion if necessary

Applies to #285

### DIFF
--- a/public/pegjs/blockUserScript.pegjs
+++ b/public/pegjs/blockUserScript.pegjs
@@ -73,10 +73,6 @@ Test the generated parser with some input for PEG.js site at http://pegjs.org/on
   }
 }
 
-// Uncomment to parse an entire metadata block for PEG.js site.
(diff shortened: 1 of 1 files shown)


Martii:
Finally found a reference to the syntax in v0.8.0 of *pegjs*



`"allowedStartRules": ["start", "non-start"]`

and

`parser.parse("input", { startRule: "myStartRule" });`


Appears to work .user.js side so leaving `block` rule name uncommented for now for easier testing on http://pegjs.org/online ... will do the same for OUJS grammar file.

See also:
* https://github.com/pegjs/pegjs/issues/136#issuecomment-41465739
* https://github.com/pegjs/pegjs/commit/98ff2eb83fb3a09655880605f248d49387d1340e

PR #679, opened 2015-07-29
Localization of metadata blocks

Martii:
* Use *pegjs* synchronously... I do have a portion of the routine for asynchronous *(done first)* but *(sync)* seems safer in the event of a production failure
* Establish the `OpenUserJS` metadata block and migrate all collaboration there
* Store localized `@name` and `@description`... not currently in use but available
* Some, but not all, line lengths in affected files wrapped according to STYLEGUIDE.md
* If no script description don't create the `p` tag on script lists
* Some stray trailing commas removed
* Some string constants shortened for error messages.. type 400 is Bad Request by standards and we shouldn't need to say that again.

**NOTE**: Still needs *mongoose* DB migration otherwise all the meta values used don't show up

* Bumping project version... e.g. once the *mongoose* *(DB data)* migration occurs there is no going back to a prior project version otherwise there will be possible DB corruption/failures *(most notably the meta.js route)* in older commit HEADS

Applies to #285

---

@sizzlemctwizzle 
In  place migration *(for both dev and pro)* can't occur *(e.g. production has to be taken offline)* since the metadata routine on production would fail with older scripts that don't have the meta stored in the new format... let me know how it goes.

For everyone else... please comment if something should be done differently *(in this methodology and not some other that has already been discussed)*... this isn't a total optimization round so please keep this in mind.


controllers/scriptStorage.js
```diff
@@ -429,6 +443,8 @@ exports.storeScript = function (aUser, aMeta, aBuf, aCallback, aUpdate) {
       }
 
       aScript.save(function (aErr, aScript) {
+        // WARNING: No error handling
```
Martii: btw notice this! I ran into an error server side and had no idea where the error was coming from... might be nice to handle this at some point.

Martii:
Uggh gimme a moment to resolve conflicts

Martii:
I'm going to close this and restart over on a fresh PR... too many issues.

PR #699, opened 2015-08-04
Apply sort parm to URI for Graveyard

### LINKED ISSUE #526: `orderBy`, `OrderDir`, and `q` QSP's not maintained in the Graveyard
1. Visit https://openuserjs.org/mod/removed
2. Click any sort method in the table header
3. Click any filter method in the sidebar
4. Voila the QSP if foo'd... same in reverse for items 2 and 3

e.g. search and sorting in the Graveyard QSP *(Query String Parameter) needs to be initialized and dynamically altered depending on view.

Similar to #181
Keep in mind #490

Martii:
* Done previously in #654 for public sortings
* Add some *font-awesome* icons to the graveyard
* Transform "All Types" to the correct nomenclature of "Clear Filters"
* Change `href` to use parent plus `/removed` e.g. `./removed` static
* Made User removal reason virtual instead of hard-coded... useful for Search filters e.g. reason is now blank when system generated... still don't allow a blank reason from a real author.. post fix for #261 and pull #513. This allows no reserved phrases other than string empty.
* Some STYLEGUIDE.md conformance
* DB remove model **migrated** as far back as "User removed" existed
* Added some common needed Filters and reordered similar to GH... not all are implemented yet

Closes #526

### DIFF
--- a/controllers/remove.js
+++ b/controllers/remove.js
@@ -44,9 +44,9 @@ exports.rm = function (aReq, aRes, aNext) {
       });
     }
 
-    // Simple error check for string null and reserved phrase
+    // Simple error check for string null
     reason = reason.trim();
-    if (reason === '' || /^User removed$/i.test(reason)) {
+    if (reason === '') {
       return statusCodePage(aReq, aRes, aNext, {
         statusCode: 403,
         statusMessage: 'Invalid reason for removal.'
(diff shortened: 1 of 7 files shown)


Martii:
That's interesting the short GH hash got longer all by itself...

`Welcome to OpenUserJS.org (OUJS) master v0.1.9+29d8f129`... 8 digits instead of 7 ... noticed that GH itself has been using longer short hashes too. *(10 currently)*

Issue #490, opened 2014-12-08
Removed items aren't fully searchable


Martii:
**EDIT**: Isolated searches are now not searchable because we needed a general search

~~As mentioned in #489 the removed items fail to search on at the very least a Remover name.~~

Some dev test points:
- [http://localhost:8080/mod/removed?q=GM_set](http://localhost:8080/mod/removed?q=GM_set) Returns nothing but it should
- [http://localhost:8080/mod/removed?q=Marti](http://localhost:8080/mod/removed?q=Marti) Returns nothing but it should

One possible resolution suggested from @jerone with:

> A found ... reason: the column `Item` is not a searchable property. The column `Type` is for example a searchable column (http://localhost:8080/mod/removed?q=user)

**EDIT**:
Searching should try to include more columns not just ones that are linked.

~~Keep in mind #526~~

Martii:
This issue is still open because it would nice to have another QSP, maybe `model` that would just return results from the items from the model removed and ignore everything else including `reason`... like it was doing before #699 ... but we still need a generalized search which is why just a `q=` is general.

Martii:
And perhaps a "not" flag QSP... to invert the search.

Martii:
Added tracking upstream label due to `content.*` not working currently.

PR #700, opened 2015-08-05
Reestablish the model filters

Martii:
* This is about the medium average point between methodologies of Zren, Jerone, SizzleMcTwizzle and myself that can be managed at the moment... I tinkered with naming the QSP `filterBy`... Item, Reason, Removed By, and Date will never be filtered imho so I ended up with a merge of `byModel`.
* Fixed a bug on `exports.`...
* Renamed the column from `Type` to `Model` since that's probably what is going to be in here... although `strategy` model probably won't but the `orderBy` QSP is already using `model`

Applies to #490

### DIFF
--- a/controllers/moderation.js
+++ b/controllers/moderation.js
@@ -66,6 +66,9 @@ exports.removedItemListPage = function (aReq, aRes, aNext) {
     });
   }
 
+  //
+  options.byModel = aReq.query.byModel !== undefined ? aReq.query.byModel : null;
+
   // Page metadata
   pageMetadata(options, 'Graveyard');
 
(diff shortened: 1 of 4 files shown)


libs/modelQuery.js
```diff
@@ -125,7 +125,7 @@ var parseUserSearchQuery = function (aUserListQuery, aQuery) {
     fullWordMatchFields: []
   });
 };
-exports.parseCommentSearchQuery = parseCommentSearchQuery;
```
Martii: This is already exported once... so matched it with it's function

libs/modelQuery.js
```diff
@@ -140,7 +140,7 @@ var parseRemovedItemSearchQuery = function (aRemovedItemListQuery, aQuery) {
     fullWordMatchFields: ['model']
   });
 };
-exports.parseCommentSearchQuery = parseCommentSearchQuery;
```
Martii: This is already exported twice... so matched it with it's function

controllers/moderation.js
```diff
@@ -77,8 +80,37 @@ exports.removedItemListPage = function (aReq, aRes, aNext) {
   // removedItemListQuery
   var removedItemListQuery = Remove.find();
 
+  // removedItemListQuery: byModel
+  if (options.byModel) {
+    modelQuery.findOrDefaultIfNull(removedItemListQuery, 'model', options.byModel, null);
```
Martii: I also didn't know what to put in the parm list where the `null` is... it does work if I do `/.*/` in [R70 above](https://github.com/OpenUserJs/OpenUserJS.org/pull/700/files#diff-bad434633cfbcb799fd8514a5b741160R70) however it transfers that to the QSP which is probably undesirable which is why I ended up doing the [R84 conditional above](https://github.com/OpenUserJs/OpenUserJS.org/pull/700/files#diff-bad434633cfbcb799fd8514a5b741160R84) to ensure the default to all models. Putting in `'$in [...]'` *(the dot, dot, dot meaning all the model names)* doesn't like the `findOrDefaultIfNull` routine as it puts the single quotes in the condition which can't happen... and of course take them out and the V8 JavaScript debugger throws an exception.

Issue #390, opened 2014-10-23
Add get/set/delete a script source with a script object... and use it.


Martii:
I'm going to "override" @sizzlemctwizzle on [this](https://github.com/Zren/OpenUserJS.org/commit/c31a0b09ce9cd0351325131ae4bd525e30bbc88d#commitcomment-6982884) but still keep it in line with the current HEAD deployment to keep this from being hung like it was in #150 

e.g. this is part of the needed optimization and can be used with migration routines.

Zren has a commit at https://github.com/Zren/OpenUserJS.org/commit/c31a0b09ce9cd0351325131ae4bd525e30bbc88d#-P0 that does part of this however his branch may be too murky to merge.

---

Some misc notes:
- Replace existing routines with actual usage of these functions
- At this time keep the abstraction layer between _mongoose_ and our code e.g. avoid at this time using methods in the models themselves in response to #686.

Martii:
So in your commit @Zren were you planning on not using a stream buffer when searching the metadata block(s)? Perhaps I'm missing the point today so please humor me as I'm still warming up from a much needed break in real life.

In short we should be ready to do this next but if you would like to show me how you wanted to do this to reassert what you want in a PR I'd be just about ready for that. Please limit it to this issue... and it may take me a bit more to fully comprehend everything you were looking for. One minor exception is as far as placing the getters and setters in the models please leave them abstracted from *mongoose*, etc... e.g. don't put them in the models as I would prefer not seeing a `Script` object created, even if temporary, every time we need to access these methods *(getters and setters)* ... I've barely looked into hooks with *mongoose* as well but that could be an option at a middle compromise point.

Zren:
> were you planning on not using a stream buffer when searching the metadata block(s)?

Probably? It was just easier to deal with as a single string than an array of buffers. I never did any performance testing with either case though. The github library (for github import) returned a string, so I needed a way to parse the metadata with a string and it seemed silly to convert it to a Buffer[] just so I could use the getMeta function.

TBH, the GitHub import feature isn't really needed since the user should just copy paste the code when setting up a new script.

Martii:
> I never did any performance testing with either case though.

[...]

Issue #19, opened 2013-11-07
Create style guide


cletusc:
We need consistent formatting of code and we should define it very explicitly. We can always do cleanup commits, or even have someone specifically assigned to do code cleanup.

PR #713, opened 2015-08-18
Some misc STYLEGUIDE.md conformance in `.../scriptStorage.js`

Martii:
* Misplaced `var` declarations ... closer to current ES5.x implementation for some others just misplaced
* Reordered on first use for all vars... not as critical but good for debugging
* Changed `s3` early initialization... shouldn't need to be **and if it did** it should always be `NOTE:`d with a comment
* Some newlines for white space usage... easier reading/debugging
* Any `require` statements all at top... `require` functions become cached and always stay loaded after first use... so really shouldn't be attempting to optimize these in our code... only perceived benefit is startup pm time but is loaded elsewhere so no real benefit.
* Some notation of `WARNING:` statement on incomplete error handling
* White-space trim that my editor didn't catch on a previous commit even though it is on all the time everywhere. :\
* `WARNING:` note added for resaving script model object for no apparent reason... again should be notated.
* Max line length of 100 conformance

Applies with #285, #486, #390, and #19 ... perhaps some more... very minimal change in project behavior here which is why it is isolated.

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -10,6 +10,7 @@ var AWS = require('aws-sdk');
 
 var Script = require('../models/script').Script;
 var User = require('../models/user').User;
+var RepoManager = require('../libs/repoManager');
 
 var cleanFilename = require('../libs/helpers').cleanFilename;
 var findDeadorAlive = require('../libs/remove').findDeadorAlive;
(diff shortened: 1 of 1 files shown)


controllers/scriptStorage.js
```diff
@@ -156,7 +162,9 @@ exports.sendScript = function (aReq, aRes, aNext) {
     ++aScript.installs;
     ++aScript.installsSinceUpdate;
 
-    aScript.save(function (aErr, aScript) { });
+    aScript.save(function (aErr, aScript) {
+      // WARNING: Resaving the script... WHY?
```
Martii: @sizzlemctwizzle Cc: @Zren
Is there a particular reason why we are resaving the `Script` object here?

Seems to me this may be a logic issue if it is truly needed as the function name implies that it is just `sendScript`... possible migration artifact as well. e.g. dead code and/or undocumented migration point.

Zren: We need to save the script object it to increment the install counter.

Martii: Ahh good point that I missed... different train of thought when STYLEGUIDE'ing... thanks.

Martii: Although no error handling present if there is a connection failure.

PR #717, opened 2015-08-24
Localization of metadata blocks

Martii:
Second attempt... needed the other boogs fixed first plus found one and missed UI metas

* Use *pegjs* synchronously... I do have a portion of the routine for asynchronous *(done first)* but *(sync)* seems safer in the event of a production failure
* Establish the `OpenUserJS` metadata block and migrate all collaboration there
* Store localized `@name` and `@description`... not currently in use but available
* Some, but not all, line lengths in affected files wrapped according to STYLEGUIDE.md *(very partial in this redo)*
* If no script description don't create the `p` tag on script lists
* ~~Some stray trailing commas removed~~ Will mitigate this later
* ~~Some string constants shortened for error messages.. type 400 is Bad Request by standards and we shouldn't need to say that again.~~ Will mitigate this later
* Some DOC/UI changes to match

**NOTE**: Still needs *mongoose* DB migration otherwise all the meta values used don't show up

* Bumping project version... e.g. once the *mongoose* *(DB data)* migration occurs there is no going back to a prior project version otherwise there will be possible DB corruption/failures *(most notably the meta.js route)* in older commit HEADS

Applies to #285

---

After some slumber and rechecks the modified migration routine should be able to knock things out on production... however I will need to take it offline later this evening to do this.

### DIFF
--- a/libs/modelParser.js
+++ b/libs/modelParser.js
@@ -116,6 +116,7 @@ var parseScript = function (aScriptData) {
   var script = aScriptData.toObject ? aScriptData.toObject({ virtuals: true }) : aScriptData;
 
   // Intermediates
+  var description = null;
   var icon = null;
   var supportURL = null;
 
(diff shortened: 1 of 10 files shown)


libs/modelParser.js
```diff
@@ -127,47 +128,41 @@ var parseScript = function (aScriptData) {
...
   if (icon) {
+    // Local downmix
     script.icon45Url = icon;
   }
 
   // Support Url
-  supportURL = findMeta(script.meta, 'supportURL');
+  supportURL = findMeta(script.meta, 'UserScript.supportURL.0.value');
   if (supportURL) {
-    script.hasSupport = true;
```
Martii: Btw this was the boog I found...

libs/modelParser.js
```diff
@@ -127,47 +128,41 @@ var parseScript = function (aScriptData) {
...
-      if (htmlStub === sanitizeHtml(htmlStub, htmlWhitelistLink)) {
-        script.support = [{
-          url:  supportURL[supportURL.length - 1],
-          text: decodeURI(supportURL[supportURL.length - 1]),
-          hasNoFollow:  !/^(?:https?:\/\/)?openuserjs\.org/i.test(supportURL[supportURL.length - 1])
-        }];
-      }
+    htmlStub = '<a href="' + supportURL + '"></a>';
+    if (htmlStub === sanitizeHtml(htmlStub, htmlWhitelistLink)) {
+      script.hasSupport = true;
```
Martii: ... should have been after the sanitized test like so.

Martii:
Here is the asynchronous routine I did about two months ago for loading *pegjs* so I can delete this branch locally:

``` javascript
var parsers = null;
async.parallel([

  // Read in the UserScript pegjs grammar file and parse the content
  function (aCallback) {
    var filename = './public/pegjs/blockUserScript.pegjs';

    fs.readFile(filename, { encoding: 'utf8' }, function (aErr, aGrammar) {
        var parserUserScript = null;

        if (aErr) {
          return aCallback(aErr);
        }

        try {
          parserUserScript = PEG.buildParser(aGrammar, { allowedStartRules: ['line'] });
        } catch (aE) {
          return aCallback(aE);
        }

        aCallback(null, { UserScript: parserUserScript });
      });
  },

  // Read in the OpenUserJS pegjs grammar file and parse the content
  function (aCallback) {
    var filename = './public/pegjs/blockOpenUserJS.pegjs';

    fs.readFile(filename, { encoding: 'utf8' }, function (aErr, aGrammar) {
        var parserOpenUserJS = null;

        if (aErr) {
          return aCallback(aErr);
        }

        try {
          parserOpenUserJS = PEG.buildParser(aGrammar, { allowedStartRules: ['line'] });
        } catch (aE) {
          return aCallback(aE);
        }

        aCallback(null, { OpenUserJS : parserOpenUserJS });
      });
  }

], function (aErr, aParsers) {
  if (aErr) {
    throw aErr;
  }

  parsers = aParsers;
  exports.parsers = aParsers;
});
```

... this uses *async* to establish the parsers... I'm still a bit leery about doing that especially since it only helps startup time on VPS and gives the potential of `parsers` being null if there's a boog... would rather have dev/pro tell me straight out if there's a grammar issue.

PR #727, opened 2015-09-01
Some dep updates

Martii:
* Going to retry *mongoose* since a bug was fixed... hopefully won't have too much down time to push/start remote pro this time around

Tested on dev and local pro ... going to do a full remote pro restart


Martii:
OMG it worked this time. :)

PR #738, opened 2015-09-14
Don't call *async* twice at these locations

Martii:
* Simplify the Code from the multiple callbacks

Mentioned at OpenUserJs/OpenUserJS.org#692

**REFS**:
* [#parallel](https://github.com/caolan/async/blob/1881168d839bfda77d086c0daad833945d241bd4/README.md#parallel)
* [#waterfall...](https://github.com/caolan/async/blob/1881168d839bfda77d086c0daad833945d241bd4/README.md#waterfalltasks-callback)

---

@sizzlemctwizzle Cc: @Zren 
We've been running ~45 days in the documents view/controller with this implementation *(and already for longer than that in others)* ... unless convinced otherwise this is around ~70ms faster in rough timing tests... not to mention double nested callbacks should usually be prefixed/suffixed with `Inner` and `Outer` to avoid confusion from other contributors... this methodology works on dev and local pro and stress tested. Let me know if there are any unforeseen issues. Thanks.

---
sidebar: https://github.com/OpenUserJs/OpenUserJS.org/pull/738/files?w=iamateapot ;)

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -762,7 +762,7 @@ exports.userGitHubRepoListPage = function (aReq, aRes, aNext) {
 
   //
   var options = {};
-  var tasks = [];
+  var tasks = null;
 
   // Session
   authedUser = options.authedUser = modelParser.parseUser(authedUser);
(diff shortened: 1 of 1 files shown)


Zren:
https://github.com/OpenUserJs/OpenUserJS.org/pull/738/files?w=1

Can't see any problems with the PR. All you did was remove the function call.

I'd probably wouldn't put it in a `tasks` variable either and instead just put the array into the function call. This way it's obvious that it's not following the pattern of using `async.parallel(tasks, ...)` at the top of the function which is used everywhere else instead of possibly missing that it uses waterfall instead at the bottom.

As for the performance benefits, the request to github is probably the slowest part for building the responses for these routes. These routes could probably be pruned out since it's only used to make importing a script more complicated than a simple copy paste.

> not to mention double nested callbacks should usually be prefixed/suffixed with Inner and Outer

They're anonymous functions.

Martii:
> Can't see any problems with the PR.

Good to see and thanks for the confirmation.

> This way it's obvious that it's not following the pattern

Part of the reason why I put in the commit the document references to those is that `parallel` is actually `series` unless a timer is involved and `waterfall` is a type of `series`... so **it is following the pattern**. It's moot at this point in time as the _template.js has some issues of not explaining this clearly. A lot more will be changed when we get full ES6.

> These routes could probably be pruned out since it's only used to make importing a script more complicated than a simple copy paste.

[We've already discussed this](/OpenUserJs/OpenUserJS.org/issues/390#issuecomment-132375656) and it's off-topic in this PR. If you feel the need to create an issue stating your case please do so. I 100% guarantee **crippling** our site features with your current recommendation with get a -1 from me and most likely an Active Admin veto where @sizzlemctwizzle will decide. The GH import routine isn't that complicated with the knowledge of knowing how to interact with our dependencies.

I still value your opinion as a consultant on things but you need to stay focused and finish what you start instead of dropping bomb-shells that appear on the surface to damage OUJS. Point said please open an issue if you think you can successfully debate why we should drop GitHub importing in favor of an archaic copy/paste op that doesn't work well on all systems. Thanks.

> They're anonymous functions.

Yes but that doesn't mean we have to reuse the same identifier name thousands of times to prevent anyone else from contributing. That is bad overall and obfuscates how the code is actually working *(think readability instead of obnoxious repetition to obfuscate what is happening)*. I was thrown off initially by this and you appear to still have not learned from this *(think hungarian apps notation and all the bugs that I've found since then... with a few of my own of course... but __some pre-existing bugs were critically dangerous__. Development is at a steady pace... sometimes a little slower when the rest of the community isn't quite caught up yet and newer features will come online when they are discussed fully not just haphazardly and incomplete like some of them have been)*.

Martii:
Abandoning this PR... it is slower for sure but due to recent clarification I now agree with @Zren when he says it's not following the template... plus I'm going to do a bunch of cleanup, soon  I hope, to ensure things are template compliant. And we are starting to nest with *async* a bit... so it is a given.

Issue #641, opened 2015-06-10
Admin Feature: Show absolute flag count


sizzlemctwizzle:
Admins need to be able to see the absolute flag count (not the count that is negated by upvotes and karma). Show absolute flag count on script page, user page, script listing and user listing pages. This would allow admin to see potentially malicious scripts that are masked by upvotes, or the rare popular script turned bad.

Martii:
Isolating this to just the admins is nice... but there was a comment somewhere about letting individual flaggers see what they have flagged in case they change their mind e.g. my thoughts are showing the list/link on each user accounts homepage to the owners and also make it available to admins+... a separate listing of each script is what is being mentioned here but I think the root'ing should be on each user first.

TimidScript:
@ sizzlemctwizzle 
OUJS-1 does it in script page, only. You should consider including the metadata of the script within nodes attributes in the html listing. That way, it allows third party scripts can use it and alter the page accordingly.

Martii:
@TimidScript 
I think there is some confusion here... this issue is for administration purposes only and not to be handled in a .user.js. Very limited things should be handled in .user.js especially when it comes to management. You'll notice I only have two scripts so far and those are unit tests for Ace and the meta routine e.g. this isn't like RoR with USO where everything had to be done in a .user.js because there was a closed source gap and a learning gap due to language differences. This projects code is in JavaScript *(ES5.1)* and usually should be done within the existing code base.

In summary not every feature you are going to want will be done client side due to server constraints and other factors yet to be discussed. This particular feature should **not** be available to client side .user.js in my current opinion.

> You should consider including the metadata of the script within nodes attributes in the html listing. 

-1 unless we encounter an issue with out of sync operations... #77 and potentially #285 are better entry points for what the server shares and your comment is an expansion on #646.

Martii:
@sizzlemctwizzle 

Do we really need a count sent to the views here?

The reason why I ask is when #643 is implemented with showing everyone who flagged a script on script homepages *(and elsewhere)*... it's easy enough to count  the visible user names that show up in **Admin** Tools... **unless** you meant showing the real count in **Mod** Tools?

Martii:
Btw here is a draft of what #643 currently looks like in a users script list:

![issue-643userscriptlistdraft1 opt](https://cloud.githubusercontent.com/assets/114709/10473859/5fd7a9ec-71ed-11e5-8333-0375e4949471.png)

[...]

PR #770, opened 2015-10-17
Issue 643

### LINKED ISSUE #643: Admin Feature: Show which users have flagged something
Related to #642. Admins need a way to see who flagged something to remove users that flag things for no conceivable reason.

Martii:
Read specific commit notes since this was at least a ~~two~~ six part'er.

Applies to #643 and leads into #641

---

Tested on dev and local pro

---

A screenshot, perhaps more:
* issue-643flaggedscriptlist opt

* issue-643flaggedscript opt

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -17,6 +17,7 @@ var Script = require('../models/script').Script;
 var Strategy = require('../models/strategy').Strategy;
 var User = require('../models/user').User;
 var Discussion = require('../models/discussion').Discussion;
+var Flag = require('../models/flag').Flag;
 
 var categories = require('../controllers/discussion').categories;
 
@@ -234,8 +235,79 @@ exports.userListPage = function (aReq, aRes, aNext) {
       pageMetadata(options, ['Flagged Users', 'Moderation']);
     }
   }
-  function render() { aRes.render('pages/userListPage', options); }
-  function asyncComplete(err) { if (err) { return aNext(); } else { preRender(); render(); } }
+  function render() {
+    aRes.render('pages/userListPage', options);
+  }
+  function asyncComplete(aErr) {
+    if (aErr) {
+      aNext();
+      return;
+    }
+
+    async.parallel([
+      function (aOuterCallback) {
+        if (!options.isFlagged || !options.isAdmin) {  // NOTE: Watchpoint
+          aOuterCallback();
+          return;
+        }
+
+        // Loop through the user list
+        async.forEachOf(options.userList, function (aUser, aUserKey, aEachOuterCallback) {
+
+          // Always convert to a snapshot copy here
+          if (options.userList[aUserKey].toObject) {
+            options.userList[aUserKey] = options.userList[aUserKey].toObject({
+              virtuals: true
+            });
+          }
+
+          // Ensure reset
+          options.userList[aUserKey]._flagged = [];
+
+          // Find any flags
+          async.parallel([
+            function (aInnerCallback) {
+              Flag.find({ model: 'User', _contentId: aUser._id }, aInnerCallback);
+            }
+          ], function (aErr, aResults) {
+            var flagList = aResults[0];
+
+            if (flagList.length > 0) {
+              options.hasFlagged = true;
+
+              // Loop through the flag list
+              async.forEachOfSeries(flagList, function (aFlag, aFlagKey, aEachInnerCallback) {
+
+                // Find the user name
+                async.parallel([
+                  function (aInner2Callback) {
+                    User.findOne({ _id: aFlag._userId }, aInner2Callback);
+                  }
+                ], function (aErr, aResults) {
+                  var user = aResults[0];
+
+                  options.userList[aUserKey]._flagged.push({
+                    name: user.name,
+                    reason: flagList[aFlagKey].reason,
+                    since: flagList[aFlagKey]._since
+                  });
+
+                  aEachInnerCallback();
+                });
+
+              }, aEachOuterCallback);
+            } else {
+              aEachOuterCallback();
+            }
+          });
+        }, aOuterCallback);
+      }
+    ], function (aErr) {
+      preRender();
+      render();
+    });
+
+  }
   async.parallel(tasks, asyncComplete);
 };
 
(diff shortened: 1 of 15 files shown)


Martii:
@sizzlemctwizzle 
\*bump\*

Martii:
Misc notes *(before merging)*:
* Consider renaming `setFlaggedListInModel` to `parseFlag`...`ToModel` and move into `.../modelParser.js` ?

controllers/user.js
```diff
@@ -234,8 +234,29 @@ exports.userListPage = function (aReq, aRes, aNext) {
...
-  function render() { aRes.render('pages/userListPage', options); }
-  function asyncComplete(err) { if (err) { return aNext(); } else { preRender(); render(); } }
+  function render() {
+    aRes.render('pages/userListPage', options);
+  }
+  function asyncComplete(aErr) {
+    if (aErr) {
+      aNext();
+      return;
+    }
```
Martii: MITIGATION: Test why this error fall through is the only one... I am guessing if we have an error in the changed views we're going to 500 and not fail gracefully.

controllers/user.js
```diff
@@ -292,9 +313,27 @@ exports.view = function (aReq, aRes, aNext) {
     tasks = tasks.concat(stats.getSummaryTasks(options));
 
     //---
-    function preRender() { }
-    function render() { aRes.render('pages/userPage', options); }
-    function asyncComplete() { preRender(); render(); }
```
Martii: "Elsewhere" before

controllers/user.js
```diff
@@ -292,9 +313,27 @@ exports.view = function (aReq, aRes, aNext) {
...
+    function preRender() {
+    }
+    function render() {
+      aRes.render('pages/userPage', options);
+    }
+    function asyncComplete() {
+
```
Martii: "Elsewhere" after

Martii:
Need some input from [those questions above](/OpenUserJs/OpenUserJS.org/pull/770#issuecomment-148893492) before anyone can merge.

Additional note...we **could** eventually show Mods just the reason, but not the user, so they can make an informed decision as to flag a critically flagged script/user but I left that alone for this PR.

controllers/user.js
sizzlemctwizzle: asyncComplete should always test for an error and handle it (at the very least a 500 page). If we can catch an error earlier on then great, but this is really the last line of defense. It would be nice if we could find a way to abstract this function and put it in libs/helpers.js since I hate seeing the same logic repeated, but I realize the difficulty of doing this.

libs/flag.js
```diff
@@ -148,3 +150,51 @@ exports.unflag = function (aModel, aContent, aUser, aCallback) {
     }
   });
 };
+
+exports.setFlaggedListInModel = function (aModelName, aOptions, aCallback) {
```
sizzlemctwizzle: This is definitely the wrong module for this function. This module is supposed to be completely devoid of any front-end view logic (notice how "aOptions" or "options" are mentioned nowhere else in this file). This module defines the logic for how flagging is handled.

controllers/user.js
Martii: Kind of figured that with the missing `aNext()` and it raised an eyebrow yesterday in my background thinking... I think I'll do that in a single PR with some cleanup just on `./controllers/user.js`... I read elsewhere that ES6 is going to need `undefined` for certain functions instead of `null` *(or worse some other huge object in a bug)* so that is why I have been splitting out the `aNext()` and `return` statements... also it makes more sense not to return something that isn't there and could change down the line.

Just a misc note too *node* 4.x went LTS according to their site e.g. 5.x is around some corner.

libs/flag.js
Martii: So where should it go?

sizzlemctwizzle:
> Consider renaming

Yes, as my line comment says, the code should be moved. Create a `flag.js` controller. Remove has a model, lib, and controller. Flag and vote should be structured like this as well. So you might as well start the flag controller for this.

> Include all model requires where ever this lands

There is [code for this](https://github.com/Martii/OpenUserJS.org/blob/Issue-643/libs/remove.js#L13) already. Vote, Flag, and Remove are all actions that I implemented in that order. They are very similar so the implementations got more well thought out as I went (with Remove being the best version).

> More STYLEGUIDE.md conformance in affected functions

Nice to have, but lower priority.

libs/flag.js
sizzlemctwizzle: Answered in my comment below.

Martii:
> Create a flag.js controller.

Done locally... need to link it in though.

> Remove has a model, lib, and controller.

That appears Greek to me at the moment... what?

> Flag and vote should be structured like this as well.

Did you want me to create a base vote.js controller too?

> There is code for this already. Vote, Flag, and Remove are all actions that I implemented in that order. They are very similar so the implementations got more well thought out as I went (with Remove being the best version). I suspect that's why you're having a problem finding the right place to put the new flag function, since it's controller logic and Flag doesn't currently have a controller.

I'll have to wait and see what you say on this part once I create the controller(s).

Thanks for the feedback... more to tinker with to get this reworking. Will buzz you later.

Martii:
@sizzlemctwizzle 
Buzz, buzz, buzz! :) Now if you want me to rename the function name please give me an exact identifier please otherwise I'll just go with that function name. If I were to include the base controller in as `flag` we might have an issue as that identifier could be taken elsewhere... so I just imported the jsm function itself.

I still don't have a clue what you meant by "Remove has a model, lib, and controller.".

Did create a vote.js controller just as a placeholder to avoid future confusion perhaps. Not linked in anywhere though.

Noticed a weird `require` too and notated it... probably why it had newlines around it... seems like something I would have done early on but not sure.

Now what is the correct next step? Fully retested on dev at current PR tip since this is super fresh and that was a simple move.

sizzlemctwizzle:
> I still don't have a clue what you meant by "Remove has a model, lib, and controller.".

Remove action:
Model: models/remove.js
Lib: libs/remove.js
Controller: controllers/remove.js

Flag action:
Model: models/flag.js
Lib: libs/flag.js
Controller: all flag controller logic is currently spread out amongst multiple controllers, but I'd like you to create a controllers/flag.js file and put your function in it (the current name is fine)

Vote action:
Model: models/vote.js
Lib & Controller: all logic for voting embedded and spread out amongst multiple controllers, but you don't need to create any files until you have something to put in them

Martii:
Oh durrr... I was taking "Remove" as something you wanted me to do... not the controller itself. Sawwy.

Martii:
So did you want a different identifier name for this function e.g. signature or do you think this is PR READY?

sizzlemctwizzle:
How about getFlaggedListForContent?

Martii:
> but you don't need to create any files until you have something to put in them

Disagree due to the confusion. If I come back to a similar case where I can't figure out where to put it I may not recall this conversation.

> `getFlaggedListForContent`

I'm game for that... I thought about calling it `sizzle` since I had no clue on naming conventions and especially since it's in a new controller.... k uno momento for the identifier change. I tossed back and forth between `Model` and `Content` during this anyhow... just didn't know what was expected. Thanks.

sizzlemctwizzle:
> If I come back to a similar case where I can't figure out where to put it I may not recall this conversation.

Good point. I actually found myself unable to remember which actions lacked which files.

> I tossed back and forth between Model and Content during this anyhow

If I see Model in the name it makes me think I'm going to get (mongoose) model objects back from the function, when this function really adds more content via `aOptions`.

Martii:
Okay... done and retested *(logged out, logged in with anyone, logged in with Admin)* ... anything else before you test it with PR READY?

I've done a shiz load of testing on this function especially when I compacted it so it should be good to go. Thanks for your patience with me and "hand holding" at my request... I'm still not clear on MVC and all the local book stores have **nothing** on this style of coding... search results just aren't doing it... may have to pay a visit to a University to get a hold of a book! :)

Martii:
Auto-merging soon as retests and pouring over the code for a few more hours seems good to go.

PR #774, opened 2015-10-21
Shift `flags` to `flags.critical` in affected models

Martii:
* Project version bump... e.g. a "hump" in the DB... going back will require an earlier commit HEAD and an earlier DB
* Change everything affected
* Tiny bit of STYLEGUIDE.md compliance
* Error trap flag user... if currently seen means that a user migration didn't occur

**NOTES**
More to go...

Applies to #641

---

**IMPORTANT!!!**: Do **not** run this on local pro yet until the migration has completed... will comment on the root issue when it is started in approximately 40 minutes. This is also not optimized but a parallel transfer in the DB... there are some things that I can issue create to make it more optimized at a later time.

### DIFF
--- a/controllers/admin.js
+++ b/controllers/admin.js
@@ -474,6 +474,6 @@ exports.authAsUser = function (aReq, aRes, aNext) {
 
     aReq.session.user = user;
 
-    aRes.redirect(user.userPageUrl);
+    aRes.redirect(encodeURI(user.userPageUrl));
   });
 };
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -79,7 +79,7 @@ var setupUserModerationUITask = function (aOptions) {
 
       removeLib.removeable(User, user, authedUser, function (aCanRemove, aAuthor) {
         aOptions.canRemove = aCanRemove;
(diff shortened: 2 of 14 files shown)


controllers/admin.js
```diff
@@ -474,6 +474,6 @@ exports.authAsUser = function (aReq, aRes, aNext) {
 
     aReq.session.user = user;
 
-    aRes.redirect(user.userPageUrl);
+    aRes.redirect(encodeURI(user.userPageUrl));
```
Martii: btw this was done in #773 but didn't feel like a merge into this older branch

controllers/user.js
```diff
@@ -1481,7 +1481,7 @@ exports.flag = function (aReq, aRes, aNext) {
     if (aErr || !aUser) { return aNext(); }
 
     fn(User, aUser, aReq.session.user, function (aFlagged) { // NOTE: Inline function here
-      aRes.redirect('/users/' + username);
+      aRes.redirect('/users/' + encodeURI(username));
```
Martii: btw this was done in #773 but didn't feel like a merge into this older branch

Issue #735, opened 2015-09-04
Autolink username via the `@username` syntax miss


Martii:
[`@swole_hamster`](https://openuserjs.org/users/swole_hamster) is [showing](https://openuserjs.org/announcements/Server_Maintenance#comment-14f96c399b1) as [`@swole`](https://openuserjs.org/users/swole)_hamster ... **EDIT**: ... from [here](https://openuserjs.org/announcements/Server_Maintenance#comment-14f96c399b1).

Ref:
* https://github.com/OpenUserJs/OpenUserJS.org/commit/091ef9cb40ea716fd30c138a7fb87b2bae0bb9c7 *(miss here probably because tokens are split with regular expression look-ahead non-capture default from [here](/chjj/marked/blob/18fb6a639a4a77dd59650879bcad10d833c40067/lib/marked.js#L463) ... or try to use _marked_ lexer?)*
* https://github.com/OpenUserJs/OpenUserJS.org/commit/6d31f785ce562ccbe820e66fa2d7f85fea5f7930 *(possibly revert until a work-around or fix is found)*

---

Trace:
``` sh-session
> renderer.text()
Crowd Task Google Search Link by @swole
> renderer.text()
_hamster
```

Martii:
And another one with `info@omertabeyond.com` [here](https://openuserjs.org/scripts/vBm/Omerta_Beyond#team).

Possibly need to see if there is a look ahead and a look behind option... and during require a space before the at symbol.

sizzlemctwizzle:
To stop matching email addresses replace the regex [here](https://github.com/OpenUserJs/OpenUserJS.org/blob/master/libs/markdown.js#L85) with this: `/(?:^|\s)@([^\s\\\/:*?\'\"<>|#;@=&]+)/`

I'm not sure why the underscore is breaking the match yet.

Martii:
> I'm not sure why the underscore is breaking the match yet.

The reason is in the trace message... e.g. the package is splitting a "chunk", processing that "chunk", and we're missing it due to routine we are using with it... this is why it would be useful to have a look ahead and look behind with that package so we could see if the "chunk" needs aggregation.

sizzlemctwizzle:
To be fair. The `text` rendering method isn't one listed in the official api. I found it by going through the source of the package.

sizzlemctwizzle:
The `html` renderer method would work, but I was trying to avoid linking in code and pre blocks.

[...]

PR #789, opened 2015-10-30
Fix autolink @username bug.

sizzlemctwizzle:
Fixes the bug reported in #735.

Tested in local dev: [http://localhost:8080/discuss/735_Testing](http://localhost:8080/discuss/735_Testing)


Martii:
Passes all my quick tests... nice! :) Thanks.

+1 Will merge this in a bit.

Eventually we will want to validate this to see if the user actually exists instead of blind linking but have a few other issues to go through I think.

sizzlemctwizzle:
We can store usernames that we successfully validated on the comment model so we don't have to do a lookup for each user on every render. This will also make it easy to tie in notications in the future.

Issue #262, opened 2014-07-13
Remove redundant legacy code and improve overall consistency


sizzlemctwizzle:
My haphazard design decisions during initial development combined with the enhancements of @Zren have left the code base full of redundancies and inconsistencies. The current state of the code is borderline incoherent. I doubt anyone other than @Zren and myself have any clue how things are structured.

The solution is to remove all redundant legacy code and to make all existing code conform to the same pattern. This includes finishing the discussion software since its lack of the ability to be voted, flagged, and removed is inconsistent. The voting and flagging routines need to be modified to match removal (since that was the last one I implemented, it's the most well formed). These routes will be converted to POST routes and accessed via Ajax on the client-side.

This issue needs to be resolved before we can move forward so added the blocking label (also I'll be changing things all over the code base and we don't want to run into merge conflicts). Please refer to our [contributing documentation](https://github.com/OpenUserJs/OpenUserJS.org/blob/master/CONTRIBUTING.md#blocking) for specific information on the meaning of that label.

Although I've assigned myself to this issue, I'll be pushing a branch to this repo called `issue-262`. If anyone wants to help out, just mirror this branch on your own repo and let me know about any changes you want me to pull in the comments here.

PR #798, opened 2015-11-05
Some more conformance on Issue controller file

Martii:
* Remove redundant setting of `options.category`... this is done typically two lines down from there.
* More focus on `options.script` compared to its intermediate
* ~~Notated possible dead code with a WARNING comment... possible BUG~~ Cc: @sizzlemctwizzle and @Zren

Originally applies to #262 ... not part of https://github.com/OpenUserJs/OpenUserJS.org/issues/262#issuecomment-57592127

### DIFF
--- a/controllers/issue.js
+++ b/controllers/issue.js
@@ -260,8 +260,11 @@ exports.view = function (aReq, aRes, aNext) {
       }
 
       // Discussion
-      options.discussion = discussion = modelParser.parseDiscussion(aDiscussion);
-      modelParser.parseIssue(discussion); // WARNING: Return value never used
+      discussion = {};
+      discussion = modelParser.parseDiscussion(aDiscussion);
+      discussion = modelParser.parseIssue(discussion);
+      options.discussion = discussion;
+
       options.canClose = authedUser &&
         (authedUser._id == script._authorId || authedUser._id == discussion._authorId);
       options.canOpen = authedUser && authedUser._id == script._authorId;
--- a/controllers/issue.js
+++ b/controllers/issue.js
@@ -109,7 +109,7 @@ exports.list = function (aReq, aRes, aNext) {
     options.openIssuesOnly = open;
 
     // Script
-    script = options.script = modelParser.parseScript(aScript);
+    options.script = script = modelParser.parseScript(aScript);
     options.isOwner = authedUser && authedUser._id == script._authorId;
 
     // Category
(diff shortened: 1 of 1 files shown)


controllers/issue.js
```diff
@@ -260,8 +260,8 @@ exports.view = function (aReq, aRes, aNext) {
       }
 
       // Discussion
-      discussion = options.discussion = modelParser.parseDiscussion(aDiscussion);
-      modelParser.parseIssue(discussion);
+      options.discussion = discussion = modelParser.parseDiscussion(aDiscussion);
+      modelParser.parseIssue(discussion); // WARNING: Return value never used
```
Martii: @sizzlemctwizzle Cc: @Zren 
What's this about? Doing a model parse and not returning it's values into the object?

Zren: parseDiscussion/parseIssue/parseBlah... will only return a new object if it's still a mongoose Document (with `discussion.toObject()`). Which means the subsequent call to parseIssue won't do that since it's a normal Object. So all the assingments are applied onto the object supplied in the parameter. So no existing bug, just bad practice.

Looks like [I goofed up](https://github.com/Martii/OpenUserJS.org/commit/4ad7a1164566b49106c888b43ccda0fcca20de07#diff-2d78be05ad5a9727f53f0f0d968212bfR168) when I wrote it.

```js
var discussion = modelParser.parseDiscussion(discussionData);
discussion = modelParser.parseIssue(discussion);
options.discussion = discussion;
```

Martii: > So all the assingments are applied onto the object supplied in the parameter.

Interesting feature of *mongoose*. Thanks. I just checked if the line is needed and it is.. otherwise `options.discussion.issueCloseUrl` becomes undefined... so I'll probably add the return value for consistency in case things change down the line.

> will only return a new object if it's still a mongoose Document (with `discussion.toObject()`).

That particular function typically makes a copy of the mongoose object thus removes any linkage to *mongoose* from my experience... so did you mean "with *out*"?

Zren: https://github.com/OpenUserJs/OpenUserJS.org/blob/master/libs/modelParser.js#L346
https://github.com/OpenUserJs/OpenUserJS.org/blob/master/libs/modelParser.js#L387

>> will only return a new object if it's still a mongoose Document (~~with~~ using discussion.toObject() [to return the new object if the function exists]).

Martii: Alright... just making sure we're on the same page... I know I use a little different nomenclature from testing aspect and my software engineering history. Thank you for the clarification.

Issue #200, opened 2014-06-24
Fix UTF-8 Support


sizzlemctwizzle:
@OpenUserJs/admin @OpenUserJs/backend @OpenUserJs/frontend 
Related to #198 and a few other bugs, I just wanted to consolidate identifying and fixing a type of bug that might be common inside of OUJS. UTF-8 characters sometimes break things in production. It doesn't happen in development for some reason. It mostly has to do with encoding going wrong (or the wrong encoding function being used) during a redirect, but it might happen it other areas that have yet to be uncovered.

So far the errors I've been able to verify in production are:
- Import using GitHub (UTF-8 in GitHub path)
- Edit script description (UTF-8 in scriptname)

@Martii 
Were there any others? Any other script manipulation I tried in production seems to work.

~~**This is a blocking bug**. I will not accept any unrelated PRs, except bug fixes, until this issue is resolved.~~ **Update:** All issues that been discovered have been fixed, but there may be lingering undiscovered bugs so I'm leaving this issue open until we're confident this has been resolved. 

**Your mission (should you choose to accept it):** Use UTF-8 characters _in production_ in any way you can possibly imagine, and report back here with your results in the comments. This way we don't waste effort testing something twice and can identify the problem areas that need to be fixed.

PR #810, opened 2015-11-06
Cleanup admin controller file

Martii:
* Don't use `options` for JSON and plain text outputs instead use module helper ... faster/cleaner... this is also a Unit Test in this area as well as #249 checking
* Added watchpoint for #200 with `encodeURI`
* ~~Only time I've encounter `nil` helper function **so far**... added notation on inconsistency and since it's the only helper function used at this time in this file condensed the `require`~~
* ~~Added watchpoint for an encountered special render op... first I've seen of these so far... I know what the output looks like but unable to find a matching view for it other than it's visual reference in Admin Tools~~

**NOTES**
* ~~Questions to follow~~
* ~~One TODO to finish depending on answers~~

Originally applies to #262 ... not part of https://github.com/OpenUserJs/OpenUserJS.org/issues/262#issuecomment-5759212

### DIFF
--- a/controllers/admin.js
+++ b/controllers/admin.js
@@ -68,44 +68,6 @@ function getOAuthStrategies(aStored) {
   return oAuthStrats;
 }
 
-// Allow admins to set user roles and delete users
-exports.userAdmin = function (aReq, aRes, aNext) {
-  //
-  var options = nil(); // NOTE: Inconsistent usage of object creation here
-  var authedUser = aReq.session.user;
-
-  if (!userIsAdmin(aReq)) { // TODO:
-    aNext();
-    return;
-  }
-
-  // You can only see users with a role less than yours
-  User.find({ role: { $gt: authedUser.role } }, function (aErr, aUsers) {
-    options.users = [];
-
-    aUsers.forEach(function (aUser) {
-      var roles = [];
-      userRoles.forEach(function (aRole, aIndex) {
-        roles.push({
-          'val': aIndex,
-          'display': aRole,
-          'selected': aIndex === aUser.role
-        });
-      });
-      roles = roles.splice(authedUser.role + 1);
-      roles.reverse();
-
-      options.users.push({
-        'id': aUser._id,
-        'name': aUser.name,
-        'roles': roles
-      });
-    });
-
-    aRes.render('userAdmin', options); // NOTE: Watchpoint
-  });
-};
-
 // View everything about a particular user
 // This is mostly for debugging in production
 exports.adminUserView = function (aReq, aRes, aNext) {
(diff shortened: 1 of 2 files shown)


controllers/admin.js
```diff
@@ -86,7 +102,7 @@ exports.userAdmin = function (aReq, aRes, aNext) {
       });
     });
 
-    aRes.render('userAdmin', options);
+    aRes.render('userAdmin', options); // NOTE: Watchpoint
```
Martii: @sizzlemctwizzle Cc: @Zren 

Do either one of you know where this physical view is? The docs for *express* [res.render](http://expressjs.com/4x/api.html#res.render) *(and related [app.render](http://expressjs.com/4x/api.html#app.render))* don't seem to cover this.

controllers/admin.js
```diff
@@ -58,10 +70,14 @@ function getOAuthStrategies(aStored) {
 
 // Allow admins to set user roles and delete users
 exports.userAdmin = function (aReq, aRes, aNext) {
-  var options = nil();
+  //
+  var options = nil(); // NOTE: Inconsistent usage of object creation here
```
Martii: @sizzlemctwizzle Cc: @Zren 

Is `nil` helper really needed here? So far I haven't seen any Object prototype lines in the controllers... but I haven't looked everywhere.

controllers/admin.js
```diff
@@ -270,7 +328,7 @@ exports.adminApiKeysPage = function (aReq, aRes, aNext) {
   // strategyListQuery
   tasks.push(function (aCallback) {
     Strategy.find({}, function (aErr, aStrats) {
-      var stored = nil();
+      var stored = nil();  // NOTE: Inconsistent usage of object creation here
```
Martii: @sizzlemctwizzle Cc: @Zren 

Again... Is `nil` helper really needed here? So far I haven't seen any Object prototype lines in the controllers... but I haven't looked everywhere.

controllers/admin.js
```diff
@@ -381,7 +428,7 @@ exports.apiAdminUpdate = function (aReq, aRes, aNext) {
   });
 
   Strategy.find({}, function (aErr, aStrats) {
-    var stored = nil();
+    var stored = nil(); // NOTE: Inconsistent usage of object creation here
```
Martii: @sizzlemctwizzle Cc: @Zren 

Again... Is `nil` helper really needed here? So far I haven't seen any Object prototype lines in the controllers... but I haven't looked everywhere.

controllers/admin.js
Zren: * https://github.com/Martii/OpenUserJS.org/commit/390b095ed837760214239ddfb5897e078ea4b348#diff-d8302114d2ebccc9c7f87ed04c5c78d4R1

* https://github.com/Martii/OpenUserJS.org/commits/master/views/userAdmin.html

Martii: Ahhh so dead code from a different code migration... Thank you!

PR #811, opened 2015-11-07
Cleanup discussion controller file

Martii:
* Add a missing `return` on err handling

Originally applies to #262 ... not part of https://github.com/OpenUserJs/OpenUserJS.org/issues/262#issuecomment-5759212

### DIFF
--- a/controllers/discussion.js
+++ b/controllers/discussion.js
@@ -6,19 +6,32 @@ var isDev = require('../libs/debug').isDev;
 var isDbg = require('../libs/debug').isDbg;
 
 //
+
+//--- Dependency inclusions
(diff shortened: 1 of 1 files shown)


controllers/discussion.js
```diff
@@ -103,49 +147,64 @@ exports.categoryListPage = function (aReq, aRes, aNext) {
   tasks.push(execQueryTask(discussionListQuery, options, 'discussionList'));
 
   //---
-  async.parallel(tasks, function (aErr) {
-    if (aErr) aNext();
```
Martii: The missing `return` statement.

PR #812, opened 2015-11-07
Cleanup script controller file

Martii:
* One double to single apostrophe set
* Remove an empty task ... empty arrays, should they exist, currently work well with *async*... just means don't do anything before calling `asyncComplete` ... not quite sure why this was in here.

Originally applies to #262 ... and overlap of https://github.com/OpenUserJs/OpenUserJS.org/issues/262#issuecomment-57592127 *(preauth)*


Martii:
One found issue... fixing momentarily.

Issue #430, opened 2014-11-18
Decide which of the current `console` messages are to be production/development/debug environment


Martii:
Currently we have a bunch of `console.log`, `console.warn` and `console.err` lines in our code... decisions need to be made on which is to be applied to the `isPro`, `isDev` and `isDbg` flag tests.

Also needed are some more console messages to actually point out the flow of our MVC. Please discuss this here.

PR #820, opened 2015-11-11
Do some console ops for visual tracking in authentication routine

Martii:
* Also trying out *chalk* dep to see if the VPS pro logs can handle this without too much visual interference... nicer in development and debug modes... if not will remove.

Applies to Code Migrations, #430,  and https://github.com/jaredhanson/passport/issues/400#issuecomment-144612814

Still getting `username is taken` with *passport*@0.3.2 and *passport-github*@1.0.0
but **not** *passport*@0.2.2 with *passport-github*@0.1.5


Martii:
*chalk* dep okay in logs... appears the escape codes are stripped out in VPS logs.

PR #823, opened 2015-11-12
Update the auth deps

Martii:
* Requires DB migration... dry run took less than 5 minutes... project version bump
* Migration is necessary due to breaking change at https://github.com/jaredhanson/passport-github/commit/1a2450db6843661ffff517612fd9f1a08474c9d4#diff-f0182e1c427d42d902575ffceb9d7fb3R97
* Dev converted as well


Martii:
Actual pro DB migration completed in under two minutes... ACTIVE! e.g. don't go back to a prior DB version please otherwise any users that use GH won't be able to log in. :)

PR #830, opened 2015-11-19
Remove some legacy dead public assets

Martii:
* No .css references found with grep
* Images are pre-*bootstrap* and we use icons *(fonts)* currently in selected packages *(save for the favicons which are used in other extensions and of course the favorite icon)*
* tab-icon.jpg is removed in favor of the PNG of same 16x16 size... doubt we would need a jpeg anyhow.

Originally applies to #262


Martii:
@DoctorSnowman 
Did you need the `tab-icon.jpg` for a special application somewhere that the PNG can't be used?

PR #832, opened 2015-11-19
Some dep updates

Martii:
* *fakes3* update
* Retested *mongoose* delete op
* Mostly bug fixes with a few new features

**NOTES**
* One NPM warning on dev with:

``` sh-session
npm WARN EPEERINVALID mongodb-core@1.2.21 requires a peer of kerberos@~0.0 but none was installed.

```

... assuming previous maintainer response of https://github.com/Automattic/mongoose/issues/3362#issuecomment-140821994 since this appears to be working *(plus we don't use kerberos other than to prevent error in compiling bins)*

See those packages changelogs for details

PR #833, opened 2015-11-19
Repair so error on `npm -ls` becomes a warning

Martii:
* Added `try...catch` just in case it's a fatal error

**NOTES**

``` sh-session
problems": [

    "peer dep missing: kerberos@~0.0, required by mongodb-core@1.2.21"

],
```

... this shows up in the stdout as being the "error" along with the full compliment of packages installed ... see #832 and parent #318

*(never would have guessed this on nodejitsu... glad we aren't there anymore)*

### DIFF
--- a/controllers/admin.js
+++ b/controllers/admin.js
@@ -337,11 +337,14 @@ exports.adminNpmListView = function (aReq, aRes, aNext) {
 
   exec('npm ls --json', function (aErr, aStdout, aStderr) {
     if (aErr) {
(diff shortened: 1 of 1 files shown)


Martii:
Just a note... the "error" could also be thrown *(and did locally)* if a dependency was removed from `./package.json` but not removed from `./node_modules`... so this should cover those situations as well... for an indication that one should probably clean `./node_modules` out and start fresh. :)

PR #834, opened 2015-11-19
Add *kerberos* dep even though we don't use it

Martii:
* This removes the warning/error on `npm install`

Ref:
* https://github.com/mongodb/node-mongodb-native/blob/2db9984f5972ff1916f46f1aab025373444aa39d/README.md#diagnosing-on-unix ... not quite what I was expecting but will still need followup when we get to *node*@6.x LTS

Applies to #832 and may need further mitigation on *node* LTS upgrade next year.

Closes kcbanner/connect-mongo#203 *(not sure if this will close automatically but need the reference anyhow)*


Martii:
Hmmm that did work on closing a remote issue... kewl. :)

PR #845, opened 2015-12-06
Try bumping up the reconnect retries and interval for MongoDB

Martii:
* Every couple of days we get disconnected and the below refs offer a possible suggestion
* Doubling the default `server.reconnectTries` and quadrupling the wait time for `reconnectInterval` *(4 seconds shouldn't be too bad instead of the default of 1)*

Refs:
* http://mongodb.github.io/node-mongodb-native/2.0/api/Server.html
* Automattic/mongoose#3588


Martii:
This appears on the surface to have improved things... restarted the server to get the most recent commits in... Will run this as a longer test and perhaps halfing the `reconnectInterval` at a later date. Need a good 30ish days though.

Issue #819, opened 2015-11-08
Fix `installName` vs `installNameSlug` references


Martii:
There's a bunch of inconsistency on this... no wonder I was thrown off... this will take a bit of time.

Martii:
@Zren 
Do you know what the primary reasoning behind using the identifier extension of `Slug`... I reread everything we have on it here but seems it just appeared with the nomenclature at one point mid last year.

1. Is it an intermediate *(temporary)* variable?
2. Does it symbolize anything special compared to a non-suffixed `installName` ?
3. ???

Zren:
Install name was sometimes used with `.user.js` tacked on at the end. I wanted to differenciate from it.

https://github.com/OpenUserJs/OpenUserJS.org/blob/e593ee2a3e7a566721c75ef877002096a83a2faa/controllers/scriptStorage.js#L28
https://github.com/OpenUserJs/OpenUserJS.org/blob/e593ee2a3e7a566721c75ef877002096a83a2faa/controllers/scriptStorage.js#L71

A `slug` is what Discourse used to call a normalized string in the url. Eg: `/t/1234/stickied-topic` From a topic with an id of `12345` and a title of `Stickied Topic`.

Notice how the spaces are converted to dashes and everything is lowercased?

My use of the word slug is probably wrong since in our case since a slug is usually used in pair with an id in the url. Since a slug can have collisions. Eg: 'A title' => 'a-title'  collides with 'a-title' => 'a-title'. Also, a forum topic's title can change.

Martii:
> Install name was sometimes used with `.user.js` tacked on at the end.

So you wanted `installName` to indicate the `authorname/scriptname[.user].js` and `installNameSlug` to indicate `authorname/scriptname`... I did some web research on this and WordPress also uses this nomenclature... They treat the url, minus the domain which is mostly true in OUJS, as the "base" and then the slug on the end with human readable terms... however we have the "...Slug" treated as a base too with "base/edit`, "base/source", etc... so overall I'm not sure the term slug is applicable and a bit of a stretch... however I stopped midstream of removing "... Slug" because of its use as a final intermediate. I also saw some distinction between the trailing extension too however that is inconsistent across the controllers *(hence this issue)*... so I got rather confused as to the reasoning.

Thank you for sharing your insights and objective on this. I'll need to ponder some more on what to do next here.

[...]

Issue #432, opened 2014-11-20
Test and support new option for *express-minify* with keeping comments


Martii:
See also:
- [x] SummerWish/express-minify#15
- [x] https://github.com/OpenUserJS/OpenUserJS.org/commit/1b595bb6a986a781fa87f04cc27e31a011b81dbb
- [x] mishoo/UglifyJS2#583 _(fixed with mishoo/UglifyJS2#608 ... retested with known derivations to the UserScript metadata block including tab characters... ~~not released yet though~~ v2.6.1+)_
- [ ] mishoo/UglifyJS2#448 _(git://github.com/mishoo/UglifyJS2#harmony and git://github.com/OpenUserJs/UglifyJS2#harmony-named+cdba43c)_ *(WIP)*
- [ ] mishoo/UglifyJS2#88 _(not blocking yet)_ ... may be fixed
- [x] SummerWish/express-minify#42
- [x] SummerWish/express-minify#44
- [x] Establish the core UI elements with _bootstrap_
- [x] Establish testing methodology _(loading two instances of _UglifyJS2_ and separated out handling of package .js vs userscript/lib scripts)_
- [x] Establish permanent url usage
- [ ] Custom config _UglifyJS2_ options. _(partially started)_
- [x] mishoo/UglifyJS2#935 with PR mishoo/UglifyJS2#962
- [x] mishoo/UglifyJS2#1001
- [x] OpenUserJS metadata block option to denote possible unstable with minification but still allow... Change UI coloring and notice of "The script author suggests that minification of this script may be unstable."
- [ ] Migrate to [*terser*](https://github.com/terser-js/terser)
- [ ] #793

---

SummerWish f.k.a Breeswish was gracious enough to grant the request for enhancement and a personal thank you goes out.
avdg fixed the newline issue in the UserScript metadata block with minification.... once again thank you too.

---

OUJS refs:
- https://openuserjs.org/discuss/How_to_disable_install_with_minification_on_my_script

Martii:
[...]

PR #858, opened 2015-12-23
Allow some minification through the source urls

Martii:
* **EXPERIMENTAL**  ES5 minification via url ... Uses same routes but `.min` prefix on extension. If ES6 is encountered it should just not minify.

**NOTES**
* ES6 minification isn't supported/stable yet but will probably request an experimental branch on *express-minify* or something else
* This is on a trial basis to see how much interest and possible clobbering happens... *uglifyjs2* claims stable support with ES5 ... this will test that. Authors should recommend debugging against original source but with a notation that minified may or may not work.
* #819 followup with renaming external urls
* Debug mode of course won't use *express-minify* currently so nothing should happen on these routes.. this is being reevaluated.

* BUG fix for libraries sharing same S3 storage space when ending in a reserved extension from script name
* BUG fix for the regular expression allowing some urls without 404

Applies to #432

### DIFF
--- a/app.js
+++ b/app.js
@@ -223,6 +223,15 @@ if (minify && !isDbg) {
   app.use(minify());
 }
 
+app.use(function(aReq, aRes, aNext) {
+  if (/\.(user|meta)\.js$/.test(aReq.url)) {
+    aRes._uglifyOutput = {
+      comments: true
+    };
+  }
+  aNext();
+});
+
 app.use(lessMiddleware(__dirname + '/public', {
   render: {
     compress: false,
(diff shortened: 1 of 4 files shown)


Martii:
Some misc stats for script(s) that minify:

| script name | native | uglified | percentage |
| ----- | ----- | ----- | ----- |
| oujs - Meta View | 29881 KiB | 17332 Kib | ~41% smaller

If a script has a bunch of comments it won't minify much currently... since those are currently kept... an option in the future could be to not keep the comments and then regenerate the metadata block... just a thought.

PR #862, opened 2015-12-24
Add HTML5 Copy to Clipboard

Martii:
* New dep for HTML5 copy to clipboard
* Serve this dep statically to the DOM
* Add button for Copy raw *(native)* URL to clipboard button
* **EXPERIMENTAL** Copy minified URL to clipboard button ... same color as label for .user.js denoting experimental for now
* Tint the `input-group` class a little darker for better distinction when inputs are disabled
* BUG FIX: Disable editing in the input box for the library URL
* Add missing "X" mustache objects to accommodate all of this from #858

**NOTES**
* TODO: Finish off the jQuery notifications... having issues with *jQuery* and *bootstrap* not reassigning tooltip text dynamically... will get to this after the holidaze.
* Linux Opera Presto does **not** support HTML5 Copy to Clipboard structures... however one can still copy the text and type `// @require`
* This is needed to potentially avoid possibly hazardous flash scripts that are already in use... also Adobe deprecated Flash.
* Icon sets are limited on choices... also have some other "copy" icons which may give a better distinction but need to ponder some on this... in the meantime use **first** glyphicon from *bootstrap* and *octicons* clipboard icons
* Not sure when an `error` will occur but disable both of these if it happens... will need to have someone report if problematic.

Applies to #432


Martii:
Error routine is called in Windows Safari

``` sh-session
Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/534.57.2 (KHTML, like Gecko) Version/5.1.7 Safari/534.57.2
```

PR #872, opened 2016-01-13
Integrate file uploads into *bootstrap*

Martii:
* Ported from http://www.abeautifulsite.net/whipping-file-inputs-into-shape-with-bootstrap-3/ . Thanks to Cory LaViska
* Uses client-side *jQuery* to simulate the text label for feedback
* **Can** support multiple files however only one is currently chosen in our base code
* Replace weird `placeholder` value which is never seen and isn't in context
* Correct some previously defined attribute minimization in the view
* Add a warning note about no feedback if JavaScript is disabled.
* Fix a bug with Opera Presto based browsers *(12.16ish)* still using `x-javascript` MIME type... e.g. allow that MIME type for file uploads

**NOTE** Untested in Safari for Windows and latest Safari for Mac... should work according to all available documentation. Will test in last Safari for Windows after deployment.

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -1363,8 +1363,9 @@ exports.uploadScript = function (aReq, aRes, aNext) {
     var failUrl = '/user/add/' + (isLib ? 'lib' : 'scripts');
 
     // Reject missing, non-js, and huge files
-    if (!script || script.type !== 'application/javascript' ||
-      script.size > settings.maximum_upload_script_size) {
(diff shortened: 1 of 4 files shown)


Martii:
PASS:`Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/534.57.2 (KHTML, like Gecko) Version/5.1.7 Safari/534.57.2`

PR #873, opened 2016-01-14
Some dep changes

Martii:
* *jQuery* failure with Preview reply... unresponsive Preview button exit and full screen missing restore arrows... reverted last dep update to prior v2.1.0
* Some bug fixes


Martii:
Following toopay/bootstrap-markdown#215

PR #898, opened 2016-02-19
Some dep updates

Martii:
* *morgan* dep updates, docs and `digits` for response time
* Please read their CHANGELOGS for additional info


Martii:
Unknown failure in this clone branch on VPS... stack trace failure... works fine in backup clone.

May have to back out these two and see if it still works in this clone.

Martii:
Rebuilt *node* and updated process manager... appears okay now... strange hiccup... in sync now.

Full dep reinstall.

PR #900, opened 2016-02-20
Fix some bugs with encoding

Martii:
* Remove the patch I put in for this portion
* Create function for encoding like GH's... see helper `encoding`
* Using encodeURIComponent from #795
* Fix bug of Fork History showing "slugged" instead of native/raw
* Fix a missing model parser item for `category` virtual model
* Fix some more bugs in modelParser.js
* Transform a watchpoint with AuthAs ... see third alternative at http://stackoverflow.com/questions/15825333/how-to-reload-current-page-in-express-js with `aRes.redirect(req.get('referer'));` or *express*@4.x shortcut of `aRes.redirect('back');`... but if referer is missing will default to `/` Ref: http://expressjs.com/en/api.html#res.redirect ... this still appears to be a bug in *express* for quite some time so this is a workaround
* Add another series of watchpoints with *express*@4.x `redirect()` doing `escape` on occasion ... special handling of `redirect`.... this particular bug doesn't appear to make much sense
* Project version bump

**NOTES**
* [ ] Followup with DB migration to add `fork.utf` to existing forks otherwise they may show up as `/` *(or a list bullet)* [Array](https://github.com/OpenUserJs/OpenUserJS.org/blob/ec98e843a9e396cb2ff70f836b7103cddcb7fd63/models/script.js#L32) typed values write themselves... interesting.
* [ ] Followup with `discussion.path` fixes. *(ready)*
* [ ] If *express* fixes their redirect issue can simplify `encode` to be a smarter encode where it only encodes when needed... which is where it started but had to back that out in this PR
* Leaving GH import alone as it still works so far
* `./app.js` has one `encodeURI` in it... leaving that be for the time being until it can be retested on production... dev isn't https so no way to test here
* Not all User urls are covered in this PR *(perhaps yet)*

Applies to #819 and #262... treads on https://github.com/OpenUserJs/OpenUserJS.org/issues/262#issuecomment-57592127 among many others.

### DIFF
--- a/controllers/admin.js
+++ b/controllers/admin.js
@@ -497,6 +497,6 @@ exports.authAsUser = function (aReq, aRes, aNext) {
 
     aReq.session.user = user;
 
-    aRes.redirect(encodeURI(user.userPageUrl)); // NOTE: Watchpoint
+    aRes.redirect(user.userPageUrl); // NOTE: Watchpoint
   });
 };
--- a/controllers/discussion.js
+++ b/controllers/discussion.js
@@ -24,6 +24,8 @@ var modelParser = require('../libs/modelParser');
 var modelQuery = require('../libs/modelQuery');
 
 var cleanFilename = require('../libs/helpers').cleanFilename;
+var encode = require('../libs/helpers').encode;
+
 var execQueryTask = require('../libs/tasks').execQueryTask;
 var statusCodePage = require('../libs/templateHelpers').statusCodePage;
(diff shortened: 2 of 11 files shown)


Martii:
ON HOLD... may have a fourth option that's a little clearer and retains some original functionality.

Martii:
Abandoning for a newer PR... it should preserve more... so far so good... about 90% tested and 85% code transformed... within current code styling.

Issue #130, opened 2014-06-10
Stop using @namespace to identify scripts


sizzlemctwizzle:
It's just ugly and unnecessary. I want to wait on merging #129 before I do this.

PR #901, opened 2016-02-21
Fix some bugs with encoding

Martii:
* Remove legacy namespace in all routes ... part of #262 and #130... officially EOL as per sizzle and confirmed by me... no redirects.
* Remove the patches I put in for this portion
* Create function for encoding similar to GH's... see helper `encoding`
* Using encodeURIComponent from #795
* Fix bug of Fork History showing "slugged" instead of native/raw
* Fix category reference bug... redirect was set to the Object itself and not one of its properties
* Fix some more missing object identifiers in modelParser.js ... not all are validated yet and some may still be missing for future proofing
* Remove and Add some more watchpoints
* Flip successful "rating updated" for Groups to debug mode and related in `./libs/modelParser.js` ... these are generating too much log traffic #430

**NOTES**
* [ ] Followup with DB migration to add `fork.utf` to existing forks otherwise they may not show up next to the bullet
* Leaving GH import alone as it still works so far ... however a watchpoint added... GH web UI always `encodeURIComponent` just about everything... will need to reaffirm existing `encodeURI` later
* The term uri is inclusive to urls as previously described in #819 ... however in this context url means it could be partially encoded whereas uri means it absolutely is encoded... except for the slashes... normally those are encoded but we don't want them encoded... again following GH example. A url containing `ä` may be just that however a uri has `%E4`. You will notice that I renamed some of our identifiers from `url` to `uri`... this is on purpose to indicate that is what we are expecting.
* *express* `redirect()` uses [`Location`](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.30) and [`Content-Location`](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.14) headers. This requires all urls to be at least URI encoded as per specs... e.g. all the browsers are adding in an `escape` for the section symbol *which may be an issue with them*. So hence why the naming of `url` vs `uri`.
* The base issue of readability of the code is still there... but at least this punches out some of the bugs mentioned.
* Compression/optimization hasn't occurred yet
* No DB migration is necessary now... this includes fixing `Re:` linkage
* The smarter encoder isn't finished yet as I may be rearranging this a bit but...

**This should knock out the BLOCKING label I put on**... at least from my perspective.

Applies to #819 and #262... treads on https://github.com/OpenUserJs/OpenUserJS.org/issues/262#issuecomment-57592127 among many others.

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -26,6 +26,7 @@ var RepoManager = require('../libs/repoManager');
 
 var cleanFilename = require('../libs/helpers').cleanFilename;
 var findDeadorAlive = require('../libs/remove').findDeadorAlive;
+var encode = require('../libs/helpers').encode;
 
 //--- Configuration inclusions
 var userRoles = require('../models/userRoles.json');
@@ -87,6 +88,9 @@ function getInstallNameBase(aReq, aOptions) {
     case 'uri':
       base = encodeURIComponent(username) + '/' + encodeURIComponent(scriptname);
       break;
+
+    case 'url':
+      base = encode(username) + '/' + encode(scriptname);
 
     default:
       base = username + '/' + scriptname;
@@ -733,7 +737,11 @@ exports.webhook = function (aReq, aRes) {
     payload.commits.forEach(function (aCommit) {
       aCommit.modified.forEach(function (aFilename) {
         if (aFilename.substr(-8) === '.user.js') {
-          repo[aFilename] = '/' + encodeURI(aFilename);
+          console.log([
+            'Webhook filename:',
+            '  ' + aFilename
+          ].join('\n')); // TODO: After some data collected, reaffirm and remove this
+          repo[aFilename] = '/' + encodeURI(aFilename); // NOTE: Watchpoint
         }
       });
     });
--- a/controllers/admin.js
+++ b/controllers/admin.js
@@ -195,7 +195,7 @@ exports.adminUserUpdate = function (aReq, aRes, aNext) {
 
       // Make sure the change is reflected in the session store
       updateSessions(aReq, aUser, function (aErr, aSess) {
-        aRes.redirect(user.userPageUrl);
+        aRes.redirect(user.userPageUri);
       });
     });
   });
@@ -497,6 +497,6 @@ exports.authAsUser = function (aReq, aRes, aNext) {
 
     aReq.session.user = user;
 
-    aRes.redirect(encodeURI(user.userPageUrl)); // NOTE: Watchpoint
+    aRes.redirect(user.userPageUri);
   });
 };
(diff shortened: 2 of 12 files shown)


controllers/scriptStorage.js
```diff
@@ -733,7 +737,11 @@ exports.webhook = function (aReq, aRes) {
     payload.commits.forEach(function (aCommit) {
       aCommit.modified.forEach(function (aFilename) {
         if (aFilename.substr(-8) === '.user.js') {
-          repo[aFilename] = '/' + encodeURI(aFilename);
+          console.log([
+            'Webhook filename:',
+            '  ' + aFilename
+          ].join('\n')); // TODO: After some data collected, reaffirm and remove this
+          repo[aFilename] = '/' + encodeURI(aFilename); // NOTE: Watchpoint
```
Martii: Some data here:

`aFilename` : [`src/RFC 2606§3/RFC 2606§3 - ä %, Unit Test/RFC 2606§3 - ä %, Unit Test.user.js`](https://github.com/Martii/UserScripts/blob/master/src/RFC%202606%C2%A73/RFC%202606%C2%A73%20-%20%C3%A4%20%25%2C%20Unit%20Test/RFC%202606%C2%A73%20-%20%C3%A4%20%25%2C%20Unit%20Test.user.js) on trigger.

Refs:
1. [`0d06785`/libs/repoManager.js#L119-L120](https://github.com/OpenUserJs/OpenUserJS.org/blob/0d067859c2a49bcb971a2838111e31bde3d0de5b/libs/repoManager.js#L119-L120)
2. [https://raw.githubusercontent.com/Martii/UserScripts/master/src/RFC%202606%C2%A73/RFC%202606%C2%A73%20-%20%C3%A4%20%25%2C%20Unit%20Test/RFC%202606%C2%A73%20-%20%C3%A4%20%25%2C%20Unit%20Test.user.js](https://raw.githubusercontent.com/Martii/UserScripts/master/src/RFC%202606%C2%A73/RFC%202606%C2%A73%20-%20%C3%A4%20%25%2C%20Unit%20Test/RFC%202606%C2%A73%20-%20%C3%A4%20%25%2C%20Unit%20Test.user.js) Raw URL copied from address bar
3. `encodeURI('RFC 2606§3 - ä %, Unit Test.user.js')` := `RFC%202606%C2%A73%20-%20%C3%A4%20%25,%20Unit%20Test.user.js`
4. `encodeURIComponent('RFC 2606§3 - ä %, Unit Test.user.js')` := `RFC%202606%C2%A73%20-%20%C3%A4%20%25%2C%20Unit%20Test.user.js`

@Zren Cc: @sizzlemctwizzle 
Would either of you say there is an issue here? *(look for the comma... e.g. `%2C` and `,` ... in list item 2, 3 and 4)*

Everything appears alright [on pro](https://openuserjs.org/scripts/Marti/RFC_2606%C2%A73_-_%C3%A4_%25%2C_Unit_Test) with the JSON... but I've also picked simple values in the Userscript and something still doesn't feel right.

PR #909, opened 2016-02-26
Add QupZilla

GreenLunar:


Martii:
@GreenLunar 
Interesting browser... couple of questions though.

1. Why such an old version Greasemonkey? e.g. is it ported from that version of GM or is it the native version for 1.8.9
2. I just installed two of my scripts and neither one is working... and I know that they are very vanilla ECMAScript5 with no special `@grant`s. I'll try it under Windows and see if I have any better luck.
3. Description of Greasemonkey says "Provides support for userscripts (www.userscripts.org)"... USO is no longer around so that may be confusing.. although I did notice the hyperlink in settings.

Adding a browser that would hopefully work with ES5 is a good thing... I'll tinker some more to see if this works before merging the doc.

Martii:
Where's the debugger?? ;) :)

Martii:
Ahh found the "Console"... although not real specific on "SyntaxError: Parser error"

GreenLunar:
I don't know how to answer to you.  All I do is translating Qupzilla into my own language.  I have posted your questions in the issue tracker of Qupzilla.

Martii:
Well you've come to the right place! :) Thanks for the PR... since I was able to get at least a basic script working in it and it looks good... it's good to advertise it... plus I do like Qt based projects. KDE to the max! *(Gnome drives me batty)* :)

I will probably do some additional screenshots since the browser itself has the plugin shipped with it.

Thank you for your contribution.

Martii:
And there you go... https://openuserjs.org/about/QupZilla

GreenLunar:
> Thanks for the PR...

Thank you for accepting it!

Martii:
Thank you for bringing it to our attention. :) Unfortunately my Linux distro is one version shy on Qt *(=5.5) so I can't assist with fixing the UserScript metadata block location issue on master over there.

PR #913, opened 2016-02-27
Rename http to https

GreenLunar:


Martii:
@GreenLunar 
Always should have a commit summary including the reason behind it. e.g. breadcrumbs for all with searches. So -1 on this part.

wikipedia upmixes to https for example and that would save a client side trip... so that part is a +1

**however** some of them don't upgrade to secure:

* One of the disadvantages of https is loss of privacy for guests which is why if http is available on some sites we should let the user choose if they want secure or not. Logging in on those sites should be smart enough to go secure.
* https uses more bandwidth for the client to negotiate... so best to avoid it when possible. Being a guest visiting is one of those instances.
* Typically an issue is created first... the Establishing Owner *(sizzle)* and Active Maintainer *(that's me)* is the exception to the rule... [`master`.github/CONTRIBUTING.md#pull-request-process](https://github.com/OpenUserJs/OpenUserJS.org/blob/master/.github/CONTRIBUTING.md#pull-request-process)
* If a http link is dead and has a corresponding https... then obviously we should fix that.

Rather than feed the hysteria of the misnomer of https is better all the time it would be wise to balance it out. Secure access doesn't mean secure DNS lookup. e.g. you might have someone eavesdrop on an open wifi connection what you are viewing but because the vast majority of DNS providers around aren't secured all that potential eavesdropper has to do is follow the DNS trail and you lose the perceived privacy of https.

So in summary... two main issues... no commit summary *(and I don't mean here on GitHub issue tracker)* and secondly some of the links have a preference of http e.g. they don't use `Strict-Transport-Security` to upgrade. So -1, -1, +1 which comes out to -1 plus the veto for not following the contribution guidelines. :)

I really appreciate the effort and assistance but I'm going to have to -1 this. Due to the lack of commit summary that's an Active Maintainer veto... **which means closure of this PR and you get to do it as a new PR with the commit summary intact.** Since another collaborator burned it for everyone... issues must be created first... that way it can be discussed first... similar to parliamentary process where a motion is made, then discussion, then vote *(sometimes)*, then approval/adoption vs disproval/rejection.

I was lenient on the initial QupZilla PR because your subject sort of covered what you were doing... with further investigation I found out you were "adding a browser to the docs" which would have been a fine commit summary.

GreenLunar:
I feel flattered to receive such elaborated comment! :open_mouth:
I have taken all of the above into consideration.

Thank you.

views/includes/documents/Chromium.md
```diff
@@ -12,7 +12,7 @@ To run userscripts on Chromium, you normally need a manager extension such as Ta
 
 [githubFavicon]: https://assets-cdn.github.com/favicon.ico
 [oujsFavicon]: https://raw.githubusercontent.com/OpenUserJs/OpenUserJS.org/master/public/images/favicon16.png
-[chromiumBrowser]: http://dev.chromium.org/Home
```
GreenLunar: This site has a preference of http with optional https. -1

Comment by @Martii
Originally posted at https://github.com/GreenLunar/OpenUserJS.org/commit/df9d5b50022cdbecfeb96fcf1daca1a07459ef16#commitcomment-16372957

views/includes/documents/Chromium.md
```diff
@@ -12,7 +12,7 @@ To run userscripts on Chromium, you normally need a manager extension such as Ta
...
-[wikipediaChromium]: http://www.wikipedia.org/wiki/Chromium_%28web_browser%29
+[chromiumBrowser]: https://dev.chromium.org/Home
+[wikipediaChromium]: https://www.wikipedia.org/wiki/Chromium_%28web_browser%29
```
GreenLunar: This site has a preference of upgrading to secure. +1

This site has a preference of http with optional https. -1

Comment by @Martii
Originally posted at https://github.com/GreenLunar/OpenUserJS.org/commit/df9d5b50022cdbecfeb96fcf1daca1a07459ef16#commitcomment-16372961

views/includes/documents/Terms-of-Service.md
```diff
@@ -38,15 +38,15 @@ Within moderation we understand the occasional necessity to conserve device stor
...
 * If a Content only license is included then the [Default Code Licensing](#default-code-licensing) will be applied with licensing arbitration, but may not always be visibly indicated on a script home page.
 * Any license that prohibits derivatives is eligible for removal.
 * Non-reusable licenses, of course, may not be used.
-* [Public domain type](http://opensource.org/faq#public-domain) licenses may mean different things in different jurisdictions and therefore are not allowed and eligible for removal.
+* [Public domain type](https://opensource.org/faq#public-domain) licenses may mean different things in different jurisdictions and therefore are not allowed and eligible for removal.
 * Any attempt to unlicense a published work will be eligible for removal.
 * In the event of any code licensing dispute resolution the [Default Code Licensing](#default-code-licensing) shall prevail if previously undeclared and/or undocumented.
```
GreenLunar: 4 comments on commit df9d5b5

@Martii
Missing commit summary -1
@Martii
I won't bother doing the rest... but this shows you how we consider a PR.

@GreenLunar
No problems. Thank you very much.

@GreenLunar
To tell you the truth, I just wanted to raise this issue (if it is an issue at all), yet I did not want someone of the team members to wonder why they need to do despairing repetitive actions.

As far as I am concerned, I have done what I wanted to be done.

Martii: > yet I did not want someone of the team members to wonder why they need to do despairing repetitive actions.

That's why issues should be made first... then discuss... etc... as [I described at the PR itself first](https://github.com/OpenUserJs/OpenUserJS.org/pull/913#issuecomment-189735612). That way you won't be doing repetitive actions e.g. unnecessary work. You did find one enhancement for wikipedia so that is something you could do with a probably +1 support for merging.

Sidebar: With wikipedia links always watch that the link is language neutral too. Typically here in the US it will redirect to https://en.wikipedia.org and we shouldn't be linking to en only pages if they have their own browser language detection when possible. Same goes for Mozilla links and any others that apply.

views/includes/documents/Chromium.md
Martii: LOL Sorry about the posting on the Commits *(this repo)* instead of the Files *(your repo)*... \*makes mental note to pay more attention to GHs ever changing features\* :)

PR #917, opened 2016-02-29
A dependency downgrade

Martii:
* *mongoose* somehow is affecting https://openuserjs.org/all by not showing all discussions... was working in previous version


Martii:
Uggh false alarm... was a Userscript doing this. Apologies... \*removes that Userscript\*

PR #932, opened 2016-03-21
Some dep updates

Martii:
* Mostly bug fixes... see their CHANGELOGS
* *async* is in rc status... some possible breaking changes. As we rely on this dep a lot additional time will be needed to test before migrating
* Retested delete op


shilingluo:
uuuu

PR #933, opened 2016-03-22
Issue 390

shilingluo:


Martii:
@shilingluo 
There's a reason my branch hasn't been merged yet... I still haven't found a reason to do this without adversely affecting the current structure and a real reason to utilize this. 

We appreciate the PR but it's going to have to wait even longer... not to mention there are conflicts with the branch now.

Issue #202, opened 2014-06-24
Simple and general guide for installing user scripts


sizzlemctwizzle:
https://openuserjs.org/corner/Userscript_HOWTO_for_newbies

This user has the right idea. I'd really like to do something like this.

PR #942, opened 2016-03-31
Fill out QupZilla a bit

Martii:
* Replace `NB`'s

**NOTE**
Minimal usage of `kbd` markup for keyboard sequence emphasis but still centric on markdown

Applies to #202 and #909


GreenLunar:
I like this presentation of @QupZilla

Well done!

Martii:
Now you have the template for native .user.js support for the front-end UI... that's the whole reason why I did it... @JRinge is the real mastermind on this... just took me a bit to put his perception into my synapses correctly. :)

Issue #944, opened 2016-04-05
Server down/stalled...


Martii:
I'm unable to get into the VPS to restart it and it's spinning in a web browser. **NOTE**: This is purely a VPS issue with our provider and not the project nor the *node* configuration.

Messaged @sizzlemctwizzle Cc: @jonleibowitz

---

Last script update on local pro at 2016-04-05T12:07:05.214Z

Refs:
* https://openuserjs.org/discuss/Huge_Number_of_Downloads *(general issue on pro)*
* https://openuserjs.org/announcements/Server_Maintenance#comment-153ed665673 *(announcement on pro)*
* https://status.github.com/messages *(May be related to queue backlog and other GH related issues today... ongoing so far... slightly affecting local pro)*
* http://forums.debian.net/viewtopic.php?f=17&t=127845 *(not exactly what happened this morning but possible if it's indexing constantly... allegedly another dist-upgrade should fix according to this)*
* http://forums.debian.net/viewtopic.php?f=5&t=127832 *(kernel update breaking things... this is my guess)*
* https://www.debian.org/News/2016/20160402 *(latest intel on fixes)*
* #425
* Tampermonkeys derived code with `q.XMLHTTPREQUEST.RETRIES` in .user.js

Martii:
Found a break in the system either a hiccup on whatever is causing this... or my distro update on laptop which may have a compatible client to connect to the latest Debian... although I did try Windows too and that failed... so not entirely sure.

I have already seen a 503 with *toobusy-js* on login that is probably GH's issue... leaving this open for a more detailed investigation over the next few days. Apologies for this unscheduled outage.

Btw dist-upgrade yielded no further updates. :\

Martii:
Looks like it's still down... on laptop here.

Martii:
No word yet.

Martii:
PENDING!... got 5 "too busy's" on login... but I have now have access... we'll see how long this stays up on the VPS.

Martii:
One server restart detected... investigating.

baivong:
503 ...

Martii:
I know... it's going to take a bit to resolve this... something is chewing up memory and causing the VPS to crash... this was before the 503 addition of *toobusy-js*... I'm probably going to take the server down, do some recompilations, and see if that helps. e.g. why it's PENDING status right now.

Patience please. :)

Martii:
[...]

PR #948, opened 2016-04-08
Test if `pipe` method has a leak

Martii:
Applies to #944


Martii:
Inconclusive so far but related to #949 patch

Martii:
Piping doesn't handle S3 errors, and could cause a server trip, so needs mitigation label is currently complete on this since we error handle these now.

Issue #37, opened 2013-11-14
Graceful failure


sizzlemctwizzle:
When we encounter an error almost anywhere, we should handle it and fail gracefully.

PR #953, opened 2016-04-08
Move `git` calls to /admin route

Martii:
* Make calls explicit to minimize possible accidental server trip during certain conditions. I would have combined these but with the server hit in #944 we are teetering with ENOMEM failures
* Stray commas removed
* Make some of the output pretty ... inversely related to #37
* Reorder /admin route on most likely utilized FIFO
* Move clone identify to /admin tools as well

**NOTE**
This should fix the main /about route from 500

Applies to #944


Martii:
> This should fix the main /about route from 500

Yes it does... added bug label

PR #989, opened 2016-04-17
Cover empty installNameBase

Martii:
Applies to #944 

---

This was a known bug from when I didn't add my script name on the DOC FAQ page.

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -238,6 +238,7 @@ exports.keyScript = function (aReq, aRes, aNext) {
 
   let rUserJS = /\.user\.js$/;
   let rMetaJS = /\.meta\.js$/;
+  let rUserMetaMinJS = /\.(?:min\.)?(?:user|meta)\.js$/;
   let rJS = /\.js$/;
 
   let acceptHeader = aReq.headers.accept || '*/*';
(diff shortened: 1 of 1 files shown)


Martii:
@janekptacijarabaci

When you get a moment would you please check on PM/GM as to why https://openuserjs.org/install/Marti/.user.js triggers two requests? It does this in GM Port matched to upstream GM beta as well. I'm a little busy at the moment with OUJS and fixing critical issues. I'd really appreciate it.

Yesterday I spent a couple of hours trying to see if our code was at fault... it wasn't when I turned off GM Port... just tested this url in latest release PM with your Fork Branch and upstream GM in latest Fx release ... it does the same thing.

Thanks so much.

janekptacijarabaci:
I do not know exactly what you mean, but https://openuserjs.org/install/Marti/.user.js returns HTTP code 400 (Bad request).
And the browser is then redirected:
https://github.com/greasemonkey/greasemonkey/blob/3.7/modules/util/showInstallDialog.js#L42

Martii:
@janekptacijarabaci 
Well you hit it right on the nose... I'm not so sure that it should be redirecting in GM... theoretically it should trap the click capture then if it doesn't meet the criteria of a 200 then it should just silently do nothing... just like when GM is disabled I think.

janekptacijarabaci:
I understand it, thanks. But... The code is there for a very long time. When you say it's a problem in upstream, please create a new issue.

Martii:
> When you say it's a problem in upstream, please create a new issue.

For starters we are both "upstream" although technically peers *(e.g. across the stream)* and OUJS makes it another peer. And finally because GM hasn't always agreed with every contributors assessments they typically get blocked... and I just happen to be one of those... so I can't make an issue at GM... which is why I asked for you assistance here. When I have time it will be changed eventually in GM Port as this is definitely a bug regardless of how long of tenure the code has... the code actually isn't that old since the e10s rewrite.

janekptacijarabaci:
> hence the keyword "re-opening"

I understand, but it's common use... request.suspend(), control anything... - if not done anything: request.resume()

Ad:

``` javascript
      if (aSuccess) {
        aRequest.cancel(Components.results.NS_BINDING_ABORTED);
      } else {
        aRequest.cancel(Components.results.NS_BINDING_FAILED);
      }
```
I see no error in the console - which console do you see this error? (AFAIK, according to the documentation it should work)...

Martii:
> I see no error in the console - which console do you see this error?

Launch it from a shell and you'll see a bunch of `dump`s.

Might be getting closer to the solution ... have to retry a couple of things.

Martii:
So this did it for GM Port 3.8beta2.0.1rc0pre *(unpublished btw)*:

``` js
  rs.download(function(aSuccess, aType) {
    if (aRequest && 'script' == aType) {
      if (aSuccess) {
        aRequest.cancel(Components.results.NS_BINDING_ABORTED);
      } else {
        aRequest.cancel(Components.results.NS_BINDING_FAILED);
      }
      var browser = aRequest
          .QueryInterface(Ci.nsIHttpChannel)
          .notificationCallbacks.getInterface(Ci.nsILoadContext)
          .topFrameElement;
      browser.webNavigation.stop(Ci.nsIWebNavigation.STOP_ALL);
    } else {
      aRequest.resume();
    }
  });
```

... **and** modify [`2a2ef52`/modules/remoteScript.js#L302](https://github.com/greasemonkey/greasemonkey/blob/2a2ef526d3840a87374c2057ca3a65729780ca36/modules/remoteScript.js#L302) to read as:

``` js
aCompletionCallback(this._cancelled, aPoint || aSuccess, aPoint);
```

Notice the extra `aPoint` added *(previously missing)* when it is cancelled.

These cancel the request if it's a type of script and if it's successful do something and if it's not do something else... but always stop the browser request and pass that through the request only once.

The different `aRequest.cancel` types don't have to be done as I said above since it's ignoring it... but might be nice to have a notification by GM perhaps similar to [the stats notification](https://github.com/greasemonkey/greasemonkey/blob/2a2ef526d3840a87374c2057ca3a65729780ca36/modules/stats.js#L148-L180).

Please feel free to get this to GM ... please oh please... of course after you test it yourself. :)

janekptacijarabaci:
I don't understand this line:
`aCompletionCallback(this._cancelled, aPoint || aSuccess, aPoint);`
This function has three parameters.

This function (aCompletionCallback) has two parameters:
`function(aSuccess, aType)`

Martii:
In [this instance](https://github.com/greasemonkey/greasemonkey/blob/2a2ef526d3840a87374c2057ca3a65729780ca36/modules/remoteScript.js#L300) it's just two ... any additional parameters are considered to be `null` or `undefined` if absent. He doesn't always send all the `channel` info onto the request observer... that's where the status codes would come in handy... I actually made it into a 3 parameter *(parm)*... with the errorMessage but took that out since it's possibly more than Anthony would want to do with a notification bar. e.g. start small.

Martii:
Hmmm hang on  a second... might have goofed here.

janekptacijarabaci:
You should invite the author, arantius, to contribute...

Martii:
Yah I goofed... ignore remoteScript.js modification altogether.

> You should invite the...

Remember I'm blocked and this is pre-analysis... approaching him on his repo is the best place to present the issue but I can't do that ... long story from over a decade ago.

janekptacijarabaci:
...and can I refer this page in "my" Issue?

Martii:
Sure... it's not like we're in a dark alley. ;) :) Everyone makes mistakes but it would be good to get GMs take on this to see if there is something missed. Ideally for tracking purposes it should be upstream with the summary. They usually like very little noise *(activity)* with concise results... whereas we don't mind a little extra here. :)

janekptacijarabaci:
@Martii

See:
https://github.com/greasemonkey/greasemonkey/commit/fb36493151aa9dce6090d13f57b89dec05430cb3

Do you have any idea?

Martii:
@janekptacijarabaci 
Saw that... I missed that use case... when I have some extra time... which will be a few days... possibly moving the check earlier to allow resume may help. e.g.  Most of my diagnostics are up [there](https://github.com/OpenUserJs/OpenUserJS.org/pull/989#issuecomment-211149438). If the status code can be returned through the parms *(args)* a simple 200 check vs non-200 should do the trick.  Again I'm totally swamped right now and have over 30 meetings this week. :\

janekptacijarabaci:
@Martii 
See https://github.com/greasemonkey/greasemonkey/pull/2415

Martii:
@janekptacijarabaci 
What about 429 code? That's the main issue here I think... if greasemonkey/greasemonkey#2415 only covers >= 500 it will miss it. We typically use a "cheat sheet" from [here](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error). So 4xx codes are client errors and of course the respective 5xx and up.

janekptacijarabaci:
Yes, you're right. Thanks. The number 429 is the exception.

Martii:
Thanks @janekptacijarabaci ,

OUJS and GM Port would feel more comfortable if all 4xx codes were covered... we can *(and already do)* send 400, 403, etc. so the double hit could potentially still exist.

janekptacijarabaci:
However, it is not generally applicable. It simply means: cancel page loading if request/channel failed (some of the codes 4xx). Using this condition on servers that do not have the problem should be harmless => the browser is behaving normally (will display a relevant error page based on the error code).

Martii:
With *node* a 400 can be sent after a DB lookup e.g. been deleted, hidden, offline, authenticated, etc. So the protection is before these e.g. pre-checks. *node* is quite a bit different than standard PHP and whatever modules are loaded.

I'll give this a whirl with all my test cases hopefully next week and see if it's really necessary but the limitation will be on GM and those peers if they don't support it... presuming it's an issue.

Thanks for your insight and be in touch.

janekptacijarabaci:
It will not be my decision. I just submitted a proposal.
Thanks for factual feedback.

Martii:
@janekptacijarabaci 
Something that doesn't really make much sense to me still are [these current GM lines](https://github.com/greasemonkey/greasemonkey/blob/fb36493151aa9dce6090d13f57b89dec05430cb3/modules/util/showInstallDialog.js#L47-L54).


Putting this to plain English... if the requests *(including external resources)* are successful, then tell the observer that "The asynchronous request failed because it was aborted by some user action." e.g. could be cancelled on installation, cancelled by server network, or cancelled by client network. This may be a sign of a larger issue because if I backtrace the "success" identifier it's actually an "error" with a logical not most of the time. It does appear to be needed with GM_hitch but based off the logic flow it should be:

* If the requests *(including external resources)* are **unsuccessful**, then tell the observer that "The asynchronous request failed because it was aborted by some user action."

Does that make sense?

As far as the PR goes it still needs to cover all 4xx codes in order to "do something by handling the error" e.g. if we send a 404 GM tries again because it's out of the "lock" already with a "key"... I haven't been able to figure a way around the browser issue of reopening the channel yet but the logic identifier phrasing in GM doesn't appear to be correct at this time and is seriously throwing off logic comprehension albeit functionally working that way. e.g. inverted nomenclature logic vs physical nomenclature execution.

Basically GM should "do something by handling the error"... however a pass-through to the site handler could be agreeable as well but doesn't appear to be available at this time.

Martii:
One other observance... `aType` shouldn't be `script` if it's content type is text/html, or equivalent since it's a web page and not a script... so that's an issue in the callback.

janekptacijarabaci:
I understand your issue. But... how to do it differently and effectively (and especially robust).

AFAIK:

At this time:
https://github.com/greasemonkey/greasemonkey/blob/3.9/modules/requestObserver.js#L90
we don't know anything else (e.g. the contentType).

Additional follows:

- https://github.com/greasemonkey/greasemonkey/blob/3.9/modules/requestObserver.js#L100
=> request.suspend();

- https://github.com/greasemonkey/greasemonkey/blob/3.9/modules/remoteScript.js#L170
=> the contentType == e.g. "text/html" => aRequest.cancel(Cr.NS_BINDING_ABORTED)

- https://github.com/greasemonkey/greasemonkey/blob/3.9/modules/remoteScript.js#L183
=> error == true
because:
=> aStatusCode == Cr.NS_BINDING_ABORTED
=> Components.isSuccessCode(aStatusCode) == false;
=> !Components.isSuccessCode(aStatusCode) == true;

- https://github.com/greasemonkey/greasemonkey/blob/3.9/modules/util/showInstallDialog.js#L47-L56
=> request.resume();
because:
=> aSuccess == !error (see above)
=> aSuccess == false

Will be necessary to resolve these issues:

- Problem loading page (Unable to connect)
- All 4xx codes (IMHO: https://github.com/OpenUserJs/OpenUserJS.org/pull/989#issuecomment-239525185)
- All 5xx codes
- The url == *.user.js$ (the contentType == "text/(x|ht)ml")
- The url == *.user.js$ (the contentType != "text/(x|ht)ml")
- etc. (maybe)

PR #1010, opened 2016-05-29
Broader `downloadURL` check

Martii:
* Fix someone who did an OUJS script homepage for a `downloadURL`... should cover any other non-Userscript-source route. Previous meta check is inclusive since that's not a Userscript source route.

Post fix for #944 and #1004 change


Martii:
Hmmm not sure if I like how it drops the PR summary and only goes off of the PR.

PR #1033, opened 2016-09-26
Some dep updates

Martii:
* Please read their CHANGELOGs
* Retested delete op


Martii:
*bundler* gem update btw... no effective change detected.

PR #993, opened 2016-04-20
Change to *colors* dep

Martii:
* Really could have used a little better initial explanation for chalk/chalk#110

Not to mention the lack of nicety I just got with the lock... we don't want bad support maintainers for sure.

*colors* is longer lived anyhow.

PR #1051, opened 2017-01-21
Change *ansi-colors* to work with `inverse` correctly

Martii:
* First error, of this kind, detected via init.js
* Needed for this init.js message on pro of:

``` sh-session
Project dependency error!

 Code 1
 Command failed: bundler outdated
```

This isn't true but seems to get past this... probably an issue with gem/bundler

Post #993

### DIFF
--- a/dev/init.js
+++ b/dev/init.js
@@ -175,7 +175,7 @@ var tasks = [
 async.waterfall(tasks, function (aErr, aResults) {
   if (aErr) {
     console.error(
-      colors.red.inverse('Project dependency error!\n\n'),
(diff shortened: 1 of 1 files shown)


Martii:
For reference:

``` sh-session
$ gem list

*** LOCAL GEMS ***

...
bundler (1.14.0)
fakes3 (0.2.5)
...
```

Martii:
Resolved with:

``` sh-session
$ sudo gem update
```

Apparently a sub-dep that *bundler* uses on production installs by default was out of date.

Issue #904, opened 2016-02-24
Consider migrating away from jQuery


Martii:
jQuery's limitations are becoming more apparent and is a perceived drain on everyone's resources with very little benefit as everything can be done natively in JavaScript with minimal impact.

Some browsers with their versions are being blocked by jQuery and has a negative effect on the world wide web browsing community. Maximum compatibility has been always been a goal and that seems to be an issue in the jQuery Core now.

The poor response from that project at jquery/jquery/issues/2940, jquery/jquery#2943, and jquery/jquery.com#121 are apparent along with a few reported violations of GH's TOS with that project. We do have at least one dependency that relies upon that dependency but I assume there is something else to replace it.

This issue is not a venue for flaming but to decide long term the stability, functionality and repercussions of the changes in jQuery... and may be long term.

Cc: @jaubourg

Martii:
gnarf [wrote](https://github.com/jquery/jquery/issues/2943#issuecomment-188033959):
> https://jquery.org/conduct/reporting/ -- this is not the right venue for this issue.

@gnarf
Thank you for that linkage however this is the discussion for this now with a more reasonable effort from staffing instead of what was received over there. I have no interest in jQuery's parliamentary process when it's clear that there won't be any fairness involved. Having jQuery promoting fear and intimidation through the community on a simple question isn't what I had in mind... so we'll make the appropriate discussions.

Please feel free to pipe in specific to this issue as well as the simple questions that I asked over there that were answered with complete utter disregard for protocol.

My Corps are now in the fold as well and are considering migrating away from jQuery as well in closed session.

gnarf:
Okay, I really don't appreciate your attitude here.  I was offering you an official venue to report a problem if you really felt that someone violated our Code of Conduct. 

I don't think that we are in any way promoting fear or intimidation, and in fact it is you who is doing so in this issue.

We've always taken a conservative stance when choosing which browsers to support in jQuery, and we try to push the envelope into cleaner, smaller footprints in newer environments lately.  We just aren't in a world where supporting oldIE is that big of a concern for most web developers nowadays.  Older versions of jQuery should continue to work on those older browsers, and we would probably consider releasing hotfixes for serious enough problems in the old branches even.

[...]

PR #1065, opened 2017-03-31
Play nice with desktop keyboard modifiers

Martii:
* Fixes Ctrl+click and others

Applies to #904 and post fix for PR #1062


Martii:
Argg I hate bugs that just appear out of nowhere... now it's not playing nice with this patch but **it was working 4 days ago*... will test killing jQuery to see if it's a hack they put in to affect standard Web API usage.

PR #1072, opened 2017-04-04
Play nice with Mac desktop keyboard modifiers

Martii:
Applies to #904 and #1062


Martii:
K... this better stay working... should know in about 4ish days. GRR.

PR #1074, opened 2017-04-05
Post Opera Presto fix

Martii:
* Scoot preventing default to a later time... conditionally logic'd better

Applies to #904 and #1062


Martii:
Different test results again today... needs mitigation.

Issue #601, opened 2015-03-14
Consider Ability to Edit Comments/Discussions


Martii:
This doesn't really have a home issue and should be referenced with this issue number. It is however referenced randomly in other tickets.

My thoughts are for a very strict edit...

1. Discussion may only be renamed by the originator and perhaps a given time frame.
2. New comments can be edited without notification of being "last edited"
3. If a newer comment is added to a discussion, then prior edits should indicate that they were "last edited" with a last accessed perhaps with the same type of Base16 date hash like our other models for the `id`.
4. Optionally if we get as, forgive the candidness but it's accurate, "anal" as GH use a sha1sum for the content as well. 
5. If the originator gets removed for spam reasons, or other, the entire discussion should be removed... or the next reply author becomes the owner... this is a choice... personally I'd rather see the entire thread disappear. One of the most common private notices I gave on USO was to not to respond to spammers. To keep things on a positive note it is best just to remove all comments under that discussion assuming the originator was at fault. Obviously if a reply is spam just the spam should be addressed.
6. If a reply is removed, it might be nice to see a brief message say "comment deleted" but this is optional.
7. Issue discussions could be controlled much like the Open/Close where the owner of a script page has control to open/close and originator has close privi's... e.g. change open/close for remove... if removed issues should say that it has been removed... or we can add all of this to flagging a comment and handle it internally...

These are a few of my thoughts. Obviously it has some thought into it already but this is what would be handy imo.

In general these steps will promote healthy commenting along with healthy reediting. All of these will, in one form or another, assist in administrative decision regarding spam messages.

sizzlemctwizzle:
For the purpose of this discussion, "deleted" means the user who posted a comment deleted their own comment. It is no longer stored and doesn't appear in the Graveyard. "Removed" means that the comment was removed by moderation and it will appear in the Graveyard.

1-3 agree
4. Not necessary.
5. Remove the whole discussion if the first comment is removed.
6. If a comment is mod removed, show nothing. If the comment is deleted, show a message that it was deleted in its place.
7. Not exactly sure if I understand, but I would rather stick to the same moderation logic we use everywhere for issue discussions. If it's spam, users flag it and mods removed it.

I'd like to add:
8. You can delete your own discussion if their are no comments by others.

I'm going to assign this to myself because I'll be handling the backend, but I have no experience with bootstrap so I'd welcome help with the UI.

Martii:
[...]

PR #1102, opened 2017-05-04
Add in the UI for comment editing

Martii:
* Right now everything returns false in mustache so no one will see it... until those identifiers match true in dev
* Using some different colors... those are handled by the pre-fab `btn-` classes.
* Utilize the *bootstrap* `hidden-xs` class to only show *font-awesome* in small viewports
* ~~One warning added of possible duplicate code added... when more solid on this it will get removed... one way or the other.~~ removed entire code content within `WARNING` blocks in b9e37b42eac8ebb036ef5ca172c34a946d3648f3
* Approximated the `editLib` usage between a combination of the other two existing... this could probably use some tuning.
* Added a field to the Comment model to handle `edited` based off naming in other models

**NOTES**
Quoting here from the mastah *(@sizzlemctwizzle)*...

> I'd be much more happy to do the backend changes for #601 if I knew the
frontend UI was just waiting for me.

It's awaiting... I'd like to see how this goes backend wise... if it doesn't eventually somewhere down the line I'll probably move from synchronous to asynchronous to get this flowing better... very lengthy process.

logged-out logged-in logged-in-forceshow logged-in-forceshow-fluid

### DIFF
--- a/controllers/script.js
+++ b/controllers/script.js
@@ -246,15 +246,6 @@ var getScriptPageTasks = function (aOptions) {
 
   // Setup the flagging UI
   tasks.push(function (aCallback) {
-    // WARNING: Possible duplicated code from flagLib.flaggable
-
-      // Can't flag when not logged in or when user owns the script.
-      if (!authedUser || aOptions.isOwner) {
-        aCallback();
-        return;
-      }
-    // /WARNING: Possible duplicated code from flagLib.flaggable
-
     flagLib.flaggable(Script, script, authedUser,
       function (aCanFlag, aAuthor, aFlag) {
         if (aFlag) {
@@ -270,12 +261,6 @@ var getScriptPageTasks = function (aOptions) {
 
   // Set up the removal UI
   tasks.push(function (aCallback) {
-    // Can't remove when not logged in or when user owns the script.
-    if (!authedUser || aOptions.isOwner) {
-      aCallback();
-      return;
-    }
-
     removeLib.removeable(Script, script, authedUser, function (aCanRemove, aAuthor) {
       aOptions.canRemove = aCanRemove;
       aOptions.flags = (script.flags ? script.flags.critical : null) || 0;
--- a/controllers/script.js
+++ b/controllers/script.js
@@ -30,8 +30,8 @@ var getFlaggedListForContent = require('./flag').getFlaggedListForContent;
 var flagLib = require('../libs/flag');
 var removeLib = require('../libs/remove');
 
-var modelQuery = require('../libs/modelQuery');
 var modelParser = require('../libs/modelParser');
+var modelQuery = require('../libs/modelQuery');
 
 var decode = require('../libs/helpers').decode;
 var countTask = require('../libs/tasks').countTask;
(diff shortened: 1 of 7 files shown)


sizzlemctwizzle:
UI looks good. I approve. Now it just needs the back-end wiring.

sizzlemctwizzle:
I should mention in case I haven't clarified this elsewhere:

Types of content: User, Script/Library, Discussion/Issue, Comment, Group

Content that can be voted on: Script/Library, Discussion/Issue, Group

Content that can be flaghed: Script/Library, Comment

Content that can be removed by moderators once it reaches flagging threshold: User, Script/Library, Comment

Users inherit the total of flags on their content (Script/Library, Comment) that has been removed. This can push them above their flag threshold allowing moderators to remove a user.

Administrators can remove all types of content at any time, and are not allowed to flag anything. They are responsible for deciding if a whole Discussion/Issue or Group should be removed since more than one user can be responsible for these types of content.

Martii:
> ... in case I haven't clarified this elsewhere...

I actually had it all working on visibility with synchronous but when it came down to `_.map` that was a major pain linking it into the asynchronous functions of the ~`whateverLib...able`... ran out of energy *(and needed sleep/break)* trying to figure out the backend but I think the `_.map` needs to be earlier outside of `preRender()` and of course utilize `async.map()`.

controllers/script.js
```diff
@@ -246,11 +246,14 @@ var getScriptPageTasks = function (aOptions) {
...
   tasks.push(function (aCallback) {
-    // Can't flag when not logged in or when user owns the script.
-    if (!authedUser || aOptions.isOwner) {
-      aCallback();
-      return;
-    }
+    // WARNING: Possible duplicated code from flagLib.flaggable
+
+      // Can't flag when not logged in or when user owns the script.
+      if (!authedUser || aOptions.isOwner) {
```
Martii: **NOTES**
* Same tests here as [...blob/`703df10`/libs/flag.js#L22](https://github.com/OpenUserJs/OpenUserJS.org/blob/703df104da6b3667687da8311c7c846d76bf9fc9/libs/flag.js#L22) and [.../blob/`703df10`/libs/flag.js#L32](https://github.com/OpenUserJs/OpenUserJS.org/blob/703df104da6b3667687da8311c7c846d76bf9fc9/libs/flag.js#L32). Looks okay to remove this entire `WARNING` block since the early test is really "dumbed down" compared to the `flaggable` routine.

Martii: Tested actual on dev and behaves identical with removal... found another duplicate in `remove.js`... testing.

Martii: Additional recheck for direct route *(no UI)* found at [.../blob/`703df10`/libs/flag.js#L176](https://github.com/OpenUserJs/OpenUserJS.org/blob/703df104da6b3667687da8311c7c846d76bf9fc9/libs/flag.js#L176) and [.../blob/`703df10`/libs/remove.js#L313](https://github.com/OpenUserJs/OpenUserJS.org/blob/703df104da6b3667687da8311c7c846d76bf9fc9/libs/remove.js#L313) ... looks secure.

Issue #1110, opened 2017-05-10
The Install / Source code pages seems currently broken (S3 hanging)


tzi:
Hi OpenUserJS Team!

I just wanted to let you know that it seems impossible (for now) to : 
 * Install a new script
 * View source code
 * Update source code

![capture du 2017-05-10 10-51-19](https://cloud.githubusercontent.com/assets/415891/25890673/b1ea5072-356e-11e7-8efe-3d659f0dfb02.png)

Cheers,
Thomas.

Martii:
@tzi 
Thanks for the report... nothing in the stderr logs... but at least my changes have kept MongoDB from posting an empty script homepage.

@sizzlemctwizzle 
S3 is really irking me... will restart the server momentarily... restarted and checked... now it works... this is the issue that prompted me to do changes for the last few weeks... it's still around. :(

tzi:
Thanks for your quick response.

It's OK for me, but I don't if you want to close this issue.
So I let you do as you want! :koala: 

Cheers,
Thomas.

Martii:
@tzi
We'll keep it open for now... apparently nothing I do prevents this yet... I'll add more to it when it appears. 

I do have to take the server down for some quick distro updates though... so if it goes completely out that's why.

Martii:
Done with VPS updates... and restarted again.

Martii:
And there it goes again... absolutely nothing extra in stderr log too. :\ At least it's not a rejection from S3. This adds HOST to the label mix too. The VPS has had connectivity issues off an on but it's up to date and their feed tells me everything usually is resolved.

Will relook into *aws-sdk* package docs and see if we're missing something else.

Martii:
New *aws-sdk* dep from #1114

Server trip:
``` sh-session
2017-05-12 12:05:44.561 +00:00: S3 GET NetworkingError for reek/Anti-Adblock_Killer_Reek.user.js in the *clipped* bucket
```

Martii:
Before reworking the error trap got this *(last restart is scheduled by me when deployed)*:

``` sh-session
2017-05-12 13:22:35.032 +00:00: S3 GET NetworkingError for vulca/topraider.user.js in the *clipped* bucket
2017-05-12 13:22:35.307 +00:00: Starting application...
...
2017-05-12 13:26:07.840 +00:00: Starting application...
```

Martii:
[...]

PR #1114, opened 2017-05-12
Some dep updates

Martii:
* Please read their CHANGELOGs... however will quote *aws-sdk* particulate:

> bugfix: ReactNative: Fixes issue where binary responses were empty in iOS (e.g. s3.getObject)

Granted we're not iOS but perhaps this will apply to #1110 since this is definitely in the ball park


Martii:
Usually don't mention deployed any more but in case *aws-sdk* is observing... deployed.

PR #1119, opened 2017-05-12
Always monitor S3 for now and address server tripping

Martii:
* Hopefully this works better for a bit... seems to on dev

*NOTE*
Out of steam... and way beyond sleep entry schedule... this will need mitigation as it's very hacky right now

Applies to #1110

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -333,35 +333,44 @@ exports.getSource = function (aReq, aCallback) {
 
       })
       .on('httpHeaders', function (aStatusCode, aHeaders) {
-        if ((isDbg || process.env.MONITOR_S3_READ_ERR === 'true') && aStatusCode !== 200) {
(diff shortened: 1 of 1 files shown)


controllers/scriptStorage.js
```diff
@@ -333,35 +333,44 @@ exports.getSource = function (aReq, aCallback) {
...
-        // Possible #486 modification
-        if (isDbg || process.env.MONITOR_S3_READ_ERR === 'true') {
-          console.error(
-            'S3 GET',
-              aE.code,
-                'for', installNameBase + (isLib ? '.js' : '.user.js'),
-                  'in the', bucketName, 'bucket'
-          );
-          s3Object = null; // TODO: Mitigate
+        // This covers network errors in addition to lookup successes and failures
```
Martii: TODO: ~~successes and~~

PR #1125, opened 2017-05-14
Replicate err traps to all S3 GETs

Martii:
* **Do nothing** if `continuation`, a.k.a. continuing, is `false` and let time out on other end if needed e.g. don't get the next chunk. This should insulate against server tripping on any GET error from S3.
* Remove `bucketName` in console stderr log since it's not really needed

**NOTES**
* Tested on dev with S3 simulation of networking error and local pro regular usage
* Still need to evaluate S3 PUTs
* Have not addressed MongoDB untrapped writes yet

Applies to #1110

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -354,15 +354,14 @@ exports.getSource = function (aReq, aCallback) {
           // fallthrough
         }
       })
-      .createReadStream()
+      .createReadStream() // NOTE: Exec equivalent
       .on('error', function (aE) {
         // This covers initial network errors and lookup failures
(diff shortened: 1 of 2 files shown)


Martii:
Followup:

S3 PUTs are covered since they aren't chunked currently at [.../blob/`c38c51e84`/controllers/scriptStorage.js#L1313-L1318](https://github.com/OpenUserJs/OpenUserJS.org/blob/c38c51e8400bfb8f9ef6ec7c06a03582f84ca4f7/controllers/scriptStorage.js#L1313-L1318)

PR #1158, opened 2017-09-25
Some dep updates

Martii:
* Delete op retested
* CSS min retested


Martii:
Source absent on install button... thought this was just a dev issue but appears to be one of the deps. Reverted to original clone for diagnosis on current HEAD.

Martii:
Seems like it's an *aws-sdk* dep failure ... only present on full reinstall of all deps... confirming with secondary testing...

Martii:
Secondary testing results in *express-minify* appearing to be the culprit... reverting until further investigation.

Martii:
Source install button present on dev... pushing to pro shortly...

Martii:
Confirmed... current HEAD okay minus *express-minify* update on pro. Will investigate in a bit.

kincaider:
840d1b6

PR #1172, opened 2017-10-12
Some dep updates

Martii:
* *mime-db* is updated fairly frequently so go back to npmjs
* Delete op retested


Martii:
Misc note: ~400,000 unique IPs in *express-brute* DB since I last cleared it. That's **HUGE**.

Issue #233, opened 2014-07-01
Allow username change


sizzlemctwizzle:
Allow users to choose another unclaimed name.

Also, we should probably prevent users from registering with an email address as their username. A simple regex can catch this and we could display a message saying the their username will be display publicly.

Martii:
@Lorentz83 

Your OUJS displayed username is now `Lorentz83`... this is a one time manual deal here.

Lorentz83:
It works, thank you.

JJanz:
Ok, so I went a similar way and registered using Google account but, instead of using my name (as, for example, GreasyFork does) it's using my email. =/
I tried to add my GitHub account hoping I could take Google away as authentication but can't remove (or change) them, just add. =/
Also, as yet seen in the topic, couldn't terminate the account and get in with GitHub as there's no Terminate Account. =/
Would you also keep my email private and change username, too, ASAP? (I'm not willing to write it here for the same reason I want to change it but OpenUserJs has it along with my GitHub username seen here.) Thanks in advance.

Martii:
@JJanz

This won't be made easy for you as it's not easy for me... ;) :) *(slightly outlined at what I have to do too)*

> it's using my email.

Strange that you would refer to yourself as an "it" since **you appear to have skipped reading the huge caution box** I put in. ;).

So let me get this squared away here...
* You just created your [userscripts repo](/JJanz/userscripts) about an hour ago.
* You created your OUJS account without reading the caution box
* You want to disguise yourself by having an auth removed.
* Finally you do know that committing anything to GH requires authentication that anyone can view which is how I've found you in our DB... so this one time change in our DB is going to give you a false sense of security.

So what I can do... since I'm in a good mood... I can do the one time rename of your account to match here, however the auths you have chosen, especially your initial signup auth method, will be staying. You have to think before your do with OUJS instead of being a drone with other less evolved sites and processes.

/end giving you a hard time for not reading. ;) :)

P.S. Your username on OUJS is "JJanz" *(without the quotes of course)*

Martii:
> (or change) them, just add

Btw something else for you to read... https://openuserjs.org/user/preferences in case you want to flip to GH as your primary. I would recommend this, especially around April 20th, 2015, since it's still a bit unclear on what is actually deprecated in googles oauth2.

JJanz:
First, thanks a lot for being this fast. I really appreciate it.

[...]

PR #1174, opened 2017-10-16
Split Login and Register into two parts

Martii:
* Something @sizzlemctwizzle suggested he might want from an idea he had with a little twist from myself. I think this covers what you are describing plus a little extra.

NOTES:
* There can be more to do with this if needed.
* This shows the sanitized name as indicated in a previous discussion about it. Added a note about it in the CAUTION alert.
* If an existing user needs to change their auth method, **and they are not already logged in with their preferred**, just page back from whatever auth page there is *(say GH's login prompt)* and it will give the register screen... otherwise use the Preferences to initially add an alternate authentication to an existing account.

### DIFF
--- a/views/pages/loginPage.html
+++ b/views/pages/loginPage.html
@@ -7,6 +7,7 @@
 <body>
   {{> includes/header.html }}
   <div class="container-fluid">
+    {{^wantname}}
     <div class="row">
       <form action="/auth/" method="post" class="form-signin">
         <input type="hidden" name="redirectTo" value="{{{redirectTo}}" />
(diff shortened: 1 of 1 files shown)


Martii:
@sizzlemctwizzle 
This should alleviate #233 from being handled by Admin+ totally *(e.g. imho it can be closed and I shouldn't have to do that any more in a DB migration)*. A new user gets a chance to see/preview what username they are going to end up with and have an option to change it. This also isolates the registration attachment to a separate screen to eliminate that confusion. The wording fits within the guidelines that you mentioned to me earlier.

Martii:
If you still want an ajax listener it is preferred to be handled before calling out the form/route request using the not-existing-yet `User.search` API that you mentioned. I advise against a real time search like Groups and that many DB lookups could get messy. This minimal amount of JavaScript won't hurt those that don't like JavaScript... they'll just need to figure it out the current existing method.

PR #1188, opened 2017-10-21
 Add some fair use bling to auth views

Martii:
* Move auths down a level in the UI... easier to see what is typed on portable devices
* Add missing class when site was converted to *bootstrap*
* Fair use icons again for the bling in modern browsers... Presto doesn't do this
* Change another label for clarity and translation size... usually smaller
* Some syncing with nomenclature from prior changes
* Fill a missing icons on button

Post #1174

---

Fix path

Post #1174


Martii:
Interesting Linux Chromium/Chrome doesn't seem to do drop down images either. :\

PR #1190, opened 2017-10-21
Safari test

Martii:
* Safari always selects some option... try a "default" at the beginning to see if it is chosen

Post #1174


Martii:
Yes it does.

PR #1201, opened 2017-10-25
Destroy session completely

Martii:
* When logging out destroy the current session not only in the User model but also the session store
* Currently `maxAge` is set to expire at browser session end... client side cookie goes away at browser private data clear but sessionId in the store sticks around for quite some time. Logging out means destroy it and login again later.
* Leaving the old `delete` in for extra cautiousness... not really needed imho as it throws an error outside of it after `destroy()`

Related to #604

### DIFF
--- a/libs/modifySessions.js
+++ b/libs/modifySessions.js
@@ -62,12 +62,16 @@ exports.add = function (aReq, aUser, aCallback) {
   }
 };
 
-// Remove a session id from the user model
+// Remove a session id from the user model **and** the session store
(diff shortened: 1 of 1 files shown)


Martii:
Adding needs mitigation label for the TODO: which entails possibly utilizing the callback and error checking.

Ref(s):
* [/expressjs/session/blob/`89fd715`/README.md#sessiondestroycallback](https://github.com/expressjs/session/blob/89fd7156129210f2b0c350afcbdf226665a8328c/README.md#sessiondestroycallback)

PR #1211, opened 2017-11-01
Add clear search feedback button to inputs

Martii:
* Using vanilla JavaScript client-side again. More accurate for single inputs.
* Disable MSIE's native... will test shortly

Loosely related to post #1174 and followup for #1210


Martii:
Slight spacing issue in MSIE 11 and OS X Safari only shows after losing focus.

Martii:
Will attempt to find an alternative after slumber.

PR #1216, opened 2017-11-01
Try an alternative clear

Martii:
* This still will duplicate under IE and Edge... not much I can do other than near visual symmetry
* Corrected one error with child triggering event in Chrome/Chromium
* Corrected one styling on the fly.

NOTES:
* Will test how IE/Edge and OS X Safari like this... Linux so far okay

Post #1211 change

### DIFF
--- a/public/css/common.css
+++ b/public/css/common.css
@@ -447,17 +447,3 @@ ul.flaggedList {
   background: transparent url("/images/ua/yandex16.png") no-repeat left top;
   color: transparent;
 }
-
-// Extend *boostrap*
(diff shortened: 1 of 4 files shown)


Martii:
hmmm **is not** duplicated under IE but **is** duplicated under Edge... love the cooperation between browsers.

Seems to be working okay in Safari.

Full available OS X browsers test okay... Chrome a little sluggish on input compared to rest via VNC.

Full available Win7 browsers test okay.

Win10 Edge is only edge-case... but it still works... just M$ imposing their clear button on everyone without a disable.

PR #1218, opened 2017-11-01
Try a generic disable of the M$ pseudo-class

Martii:
* This worked in a .user.js \*eyeroll\*

Post #1216


Martii:
Seems to work.... Gehh.

Now if I could just get Safari to work properly we could utilize the original styling.

Martii:
Ask and received... OS X Safari 11.0.1 appears to work now from App Store update just now from today.

PR #1221, opened 2017-11-02
Try the original again

Martii:
* Post testing on IE/Edge and Safari

NOTES:
* Presuming this works two additional text boxes to transform with this... may rename script inclusion

Post #1211 regress merge


Martii:
So far so good... Safari 11.0.1 is needed since 11.0.0 was buggy; Edge and IE are reacting as expected; Chromium/Chrome okay with error trap still.

PR #1321, opened 2018-02-01
*Ace*

Martii:
NOTE:
* odd error that appeared on some scripts not loading ace in any browser... updating to see if it helps


Martii:
> some scripts not loading ace in any browser

Under 10 scripts had no `isLib` value e.g. `undefined` and not boolean `true` or `false`.  Manually fixed the DB to include these... usually very early initial publish dates that propagated through updates. Mostly founders but a few others around that period of time. Probable causes are either initial code implementation and/or *mongoose* dep.

Issue #1323, opened 2018-02-09
SSL/TLS failure with requests and EPROTO response


Martii:
The following error has shown up just recently and is preventing some script uploads and the online editor from accepting scripts with a server trip:

``` sh-session
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: write EPROTO 140596207068992:error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:../deps/openssl/openssl/ssl/s3_pkt.c:1500:SSL alert number 40
140596207068992:error:1409E0E5:SSL routines:ssl3_write_bytes:ssl handshake failure:../deps/openssl/openssl/ssl/s3_pkt.c:659:

    at _errnoException (util.js:1022:11)
    at WriteWrap.afterWrite [as oncomplete] (net.js:867:14)
```

I'm in the process of rebuilding the current LTS *node* to see if that will help. That particular file is part of *node* and is beyond our control other than a rebuild to see if a VPS update broke the build.

* Restored SSL keys from November 2017 *(primary)* to be sure file corruption had not occurred. No change.
* Reinit project deps e.g. `rm node_modules` and `npm install`. No change
* New temporary project clone created. No Change
* Rolling back *node* to earlier version with *nvm* did **not** help.
* Rolling back to an older project clone did **not** help.
* Rebuilding *node*... No change
* Remove *nvm* completely... pending retest of system *node* recompile... .bashrc is new location instead of .profile.... difference in reinstallation merits removal, perhaps permanently.
* Test manual removal of https://github.com/OpenUserJS/OpenUserJS.org/blob/8d90bfa0e2e742dfcbeebb580785b3044cc98351/app.js#L247-L270 and use native *node* encryption handling. No change
* Test non-secure mode... FAIL... lovely... browser issue always kicking in https. Will followup with that later.

Outside ref(s):
* https://github.com/Martii/UserScripts/commit/5180c68cd12b902aa72fac2b637d85655c23d064 *(worked fine on this day with a GH import and a manual script edit on OUJS)*
* https://openuserjs.org/garage/Problems_pushing_scripts_from_Github#comment-161779f7995 *(Report)*

Martii:
Cause Identified: Definitely *node* throwing an untrapped error and dealing with TLS.

Workaround: Remove icon request since https://confluence.bredex.de/images/icons/profilepics/default.png does this consistently so far.

Additional session failure using *request* on dev:

``` sh-session
[...]

PR #1324, opened 2018-02-09
Suspend automatic `@icon` check and resume acct removal in the interim

Martii:
* Apparently there is a long standing issue with EPROTO and this. Will need to dive in way deep and wait until it is fixed.
* Project error trapping yielded no results
* Altered to use *request* and same issue as denoted on the issue
* Prevents the server trip from *node* builtin and other mentioned package.

Applies to #1323 and post #1303


Martii:
Please note this code is **not dead just suspended**... we can't have ppl intentionally tripping the server. Rejection is much nicer than account removal for being bad with huge `@icon`s... so back to the old method until it can be retested.

PR #1335, opened 2018-03-02
Migrate `Buffer` usage

Martii:
Applies to nodejs/node#19079 and initially referenced in da49fff


Martii:
All basic testing PASS with latest *node*@8.x and *node*@9.x

PR #1338, opened 2018-03-28
Make userscript search the new default for Chrome's "Tab to search"

erlendorf:
When auto-adding a OpenSearch plugin on the first visit of a website, Chrome seems to take the first OpenSearch plugin from the HTML header. This commit changes the order of the OpenSearch plugins in the HTML header, so that userscript search is the first one. Will probably only work for first time visitors of the site.

### DIFF
--- a/views/includes/head.html
+++ b/views/includes/head.html
@@ -6,9 +6,9 @@
 <link rel="shortcut icon" href="/images/favicon.ico" >
 
 <!-- Open Search -->
+<link href="/xml/opensearch-userscripts.xml" type="application/opensearchdescription+xml" rel="search" title="OpenUserJS Userscripts">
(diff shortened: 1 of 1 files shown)


Martii:
That's goofy of them and so much for alphetization. For a visual representation of what is in the commit message here's where to find this:

<a href="https://user-images.githubusercontent.com/114709/38057754-f350951c-32cf-11e8-85de-ec0cbb66e61d.png" title="Click to enlarge" alt="Chrome search engine defaults"></a>.

1. Clean profile
2. Visit OUJS
3. Open chrome://settings/searchEngines

Not sure what will happen in an existing profile that has Groups already there but removed.... will it add again? 

No other flags found for specifying the default for it to inject without users permission for Chrome in a quick search.

+1

Martii:
Thanks and deployed.

Looks like it's "smart" enough to readd it with the new setting.

Issue #1347, opened 2018-04-18
*passport-steam* failing


Martii:
Both 1.0.8, 1.0.9 and the OUJS forks *(commit specified)* are producing `?authfail`.

Martii:
Still failing when deployed via `git`... not sure what to do and I'm outta time for now.

Martii:
Some notes out loud *(may add/change to this)*:

1. This **is directly related** to the non-user profiled session found at #1344 earlier today ... tested on dev and it creates one on auth fail. `sessionStore.clear();` has no effect when the "stuck" session data is removed *(other than logging out everything by removal of the MongoDB session cookies)*.
2. `aLoggedIn` argument at https://github.com/OpenUserJS/OpenUserJS.org/blob/193c04c872473cc0d65e82a63941a2fa2676ecb8/libs/passportVerify.js#L16 shows as `undefined` e.g. blank in console output... which triggers this section at https://github.com/OpenUserJS/OpenUserJS.org/blob/193c04c872473cc0d65e82a63941a2fa2676ecb8/libs/passportVerify.js#L75-L77 which in turns triggers https://github.com/OpenUserJS/OpenUserJS.org/blob/193c04c872473cc0d65e82a63941a2fa2676ecb8/controllers/auth.js#L242-L248
3. Confirmed that https://github.com/OpenUserJS/OpenUserJS.org/blob/193c04c872473cc0d65e82a63941a2fa2676ecb8/libs/passportLoader.js#L37 is set to match docs at [/liamcurry/passport-steam/blob/`92974ba`/README.md#configure-strategy](https://github.com/liamcurry/passport-steam/blob/92974ba3bd6042b4d714e91db6f59412f0804a6e/README.md#configure-strategy)
4. https://github.com/OpenUserJS/OpenUserJS.org/blob/193c04c872473cc0d65e82a63941a2fa2676ecb8/controllers/auth.js#L195-L199 has the right number of parameters according to docs at [/liamcurry/passport-steam/blob/`92974ba`/README.md#configure-strategy](https://github.com/liamcurry/passport-steam/blob/92974ba3bd6042b4d714e91db6f59412f0804a6e/README.md#configure-strategy)
5. Reverted way back to 9688002 in dev... still failing e.g. probable that it's steam site related and/or the dependency *(however since *passport-steam*@1.0.8 doesn't work... it's possibly site related instead)*

Martii:
@sizzlemctwizzle 

[...]

PR #1354, opened 2018-04-23
Put in a mult-auth collision check

Martii:
* Best to be safe on this in case another auth is present and for some way rare case it has the exact same digest. Can handle these manually if need be but should be extremely rare.

Applies to #1347


Martii:
Okay reallllllllllly tired here... hang on a second...

PR #1390, opened 2018-06-01
GDPR2

Martii:
* Remember and recall consents
* Speeds up logging in if we already have it

NOTES:
* Good thing I twiddled with this before :)

Post #1174 and #1385 'ish


Martii:
Booted everyone off so this can take effect *(apologies but necessary)*. If we don't have it then users need to check it and it should remember/recall it for future logins.

PR #1399, opened 2018-06-06
Some GH webhook init events trapped with status codes

Martii:
* Informational

NOTE:
* Not tested yet as this is a pro thing from here... will test shortly
* There will be more


Martii:
Test pro... PASS on GH with 202 and PASS on OUJS ... webhook stable

PR #1423, opened 2018-06-14
Strategies update

Martii:
* Allow GitLab possibility... set to ReadOnly *(RO)* until tested
* Set all other strategies that aren't initialized yet to RO e.g. the ones that don't show up in the dropdown. This is useful for testing first on dev/local pro to see if things are going to work first before we actually commit to allowing users to use it
* New dep is based off of *passport-facebook* and appears to contain the same values. This is OAuth2.
* Added Fair Use icon for GitLab strat
* @sizzlemctwizzle will probably need to establish and authorize an account for our app access on an account there *(one for testing perhaps and one for local/production)*

Related to #127 and a much smaller step first.

NOTES:
* This is probably a good replacement for the lost *passport-aol* and is another SCM
* ~6 days ago the v3 API login failed... soooooo might still be somewhat unstable but won't know for sure until availability and testing occurs
* GitLab doesn't appear to have a solid, and dedicated, IP range yet:

Refs:
* https://gitlab.com/gitlab-com/infrastructure/issues/434
* https://gitlab.com/gitlab-com/infrastructure/issues/1985
* https://gitlab.com/gitlab-com/support-forum/issues/2966

... without these a webhook will probably never happen *(and I don't look forward to maintaining ever changing IPs from a forum thread)*. It's been almost 4 years now since the original issue was opened.

PR #1424, opened 2018-06-14
Try naming this correctly

Martii:
Post #1423


Martii:
Hmmm that didn't work in dev either... pro is offline for the moment

Martii:
Back online... note to self when tired... try on dev first. Apologies

PR #1427, opened 2018-06-14
GDPR10A

Martii:
* Occasionally there is a race condition the browser detects with redirects... so do this with a statusCodePage instead


Martii:
Very unusual auth bug... the 502 stops the race condition however it may land on that page a few times before successful retries.

Martii:
Last option if something breaks is to do just `aNext()` and ignore the error... that should theoretically return a 404 which probably wouldn't be helpful.

PR #1444, opened 2018-06-24
GDPR12A

Martii:
* Try setting `strict` on `expand`... found one extension that tells me that it's client side... so should work.

NOTE:
* Booting everyone off when this gets deployed shortly

---

Post #1439


Martii:
Deployed and booted to propagate... apologies but that's why I usually do things like this late EST.

PR #1469, opened 2018-06-29
More login checks

Martii:
Post #1468 #1467 #1466 ... related to #1446


Martii:
Think that's all of them.

PR #1473, opened 2018-06-30
Revert to TTL

Martii:
* The cost of `interval` mode is too high plus it also has a similar, yet muffled, issue as `native`. Each session gets it's own server side *(OUJS)* timer and that can be really CPU intensive if too many users pop on board at the same time.
* MongoDB remote is not cleaning up expired sessions until many, many, many hours later... this may subside at a later date but something we need to live with at the moment until the main DB is migrated to a newer version *(probably one minor semver at a time)*. \*hint hint\* @sizzlemctwizzle
* Reoutput the `originalMaxAge` as a label.
* New dep extension *(plugin as they call it)* that extends *moment*
* Max extension age is n *(currently 6)* times 3... may change these values after some more consultation

Post #1471 ... related to #604


Martii:
Needs mitigation... unpredictable results again. Trying new deps in #1477 ... could write a single timer event to auto clean similar to `interval` mode however current TTL didn't time me out on extension. Did when it was 6 but at times 3... so need to wait for 18 times 3... unfortunately won't be available to scan at dev station.

PR #1475, opened 2018-07-01
TLS session resumption enable

Martii:
Ref:
* https://nodejs.org/api/tls.html#tls_event_resumesession


Martii:
That's curious... seems to only work fully on one core. Session resumption caching is disabled. Will need to look into this some more.

PR #1483, opened 2018-07-06
Abstract TTLs to settings Object

Martii:
* Adjust values to shorter, preferred, intervals
* Additional sanity check for idle session from browser blocked and query accept redirect with JavaScript disabled.
* Restore #1448 *(not found in stderr and new users are okay)*... ends mitigation from #1449 ... most likely a critical DB failure with mongolabs that should be trapped and handled
* Move destruction to stderr with debug mode only... Applies to #430

NOTE:
* Unit specific tests... if changed must keep in `m` and `h`. `nominal` replacement and `maximum` addition are in `h`... the rest are `m`

Post #1471 ... related to #604

### DIFF
--- a/app.js
+++ b/app.js
@@ -106,7 +106,7 @@ db.on('connected', function () {
     console.log(colors.green('Connected to MongoDB v' + aInfo.version));
   });
 
-  ttlSanityTimer = setInterval(ttlSanity, 15 * 60 * 1000); // NOTE: Check every n minutes
+  ttlSanityTimer = setInterval(ttlSanity, settings.ttl.timerSanityExpiry * 60 * 1000); // NOTE: Check every n min
 });
 
 db.on('disconnected', function () {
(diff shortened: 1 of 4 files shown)


Martii:
Explaining the additional sanity check to the masses... Here's a a temporary modification at https://github.com/OpenUserJS/OpenUserJS.org/blob/6e50990cc6b5e93c407e22e92129038bbb66415a/controllers/auth.js#L334 with a fragment added of `+ #loggedin` in Chromium based browsers. Note the lack of Username in the upper left hand corner. The redirect in the browser itself is doing something weird and makes it look like one isn't logged in... but in reality one is actually logged in. The additional sanity check automatically destroys idle, abandoned, sessions like these. 

Screenshot from dev *(same in production)*:
* <a href="https://user-images.githubusercontent.com/114709/42359847-ffd7a288-80a1-11e8-99e9-0d900923f61f.png">logged in but not visible</a>

Martii:
Blehh.... now Fx 61.0.1 is doing it in a clean profile. GRR!

PR #1524, opened 2018-09-24
Migrate to *terser* dep

Martii:
* Move `console` output to dev only for experimental minification

Applies to #432 #249 #430


Martii:
NOTE: *terser* is still split into two due to experimental minification status. If and when this goes out may combine them however if a newer ESnth minifier comes out this would still allow it.

Issue #1526, opened 2018-10-08
Google+ deprecation


Martii:
Here we go again??? 

Refs:
* https://www.ghacks.net/2018/10/08/google-shuts-down-google/
* https://blog.google/technology/safety-security/project-strobe/
* Query on passport at jaredhanson/passport-google-oauth2#50

In the short term we should probably disable new Google auths @sizzlemctwizzle

Historical refs:
* #889
* #619  Google+ change... passport example uses `profile` vs. what was done here. e.g. no path.
* #484 
* peer dep fragment issue tracker at https://github.com/jaredhanson/passport-google-oauth2/issues

Ref:

Over at PR No. 51 on [the package issue tracker](https://github.com/jaredhanson/passport-google-oauth2/pulls) it states:

>#### Google Plus API Deprecation
> The default for `passport-google-oauth2` is to use the Google+ People API to
> get the profile information for the user signing in.  With this API being
> removed in early 2019 you should look towards using the OAuth 2 User Info
> endpoint instead.  You can switch to this endpoint using this configuration
> option.
```javascript
new GoogleStrategy({
  clientID: GOOGLE_CLIENT_ID,
  clientSecret: GOOGLE_CLIENT_SECRET,
  callbackURL: "http://www.example.com/auth/google/callback",
  // This option tells the strategy to use the userinfo endpoint instead
  userProfileURL: "https://www.googleapis.com/oauth2/v3/userinfo",
}
```
> The structure of the `profile` object will be the same but you may get
> less profile information than you did before.

... seems to also fix OpenID which I was told and read by Goo that it was deprecated. We took action upon that in #484 ... a trite bit confusing to see [that fix under a file named openid](https://github.com/jaredhanson/passport-google-oauth2/pull/51/files#diff-403bab8ce16e512d0f9c5dc48404eb95R26)

Martii:
Successful login to OUJS with no changes to source. Perhaps they haven't done it yet?... or perhaps that alias someone was talking about. I will check again periodically throughout the day.

Martii:
Okay... definitely past the deadline... still able to log in *(and observed at least two others with goo logging in successfully)*. I did remove g+ from my other account before all of this. 

Guess either the `profile` commit or their alias is in effect. Really don't like guessing what is happening here. 

Sometime next week I'll retry allowing new account creations.

Martii:
Interesting read at https://www.bbc.com/news/technology-47771927 ... Tuesday is their EOL?

Issue #889, opened 2016-02-04
Migrate from *passport-google-oauth* to *passport-google-oauth20*


Martii:
Same dealer ;) slimmer packaging...
- https://github.com/jaredhanson/passport-google-oauth2
- https://www.npmjs.com/package/passport-google-oauth20

... and the cost so far in a dev try branch...
- It almost kicked me out my google+ account... which didn't thrill me... which it said ~"Something's different about you..." ... nothings different... still the same 'ol dev station ... enjoying the refresher of why I have my own issues with goo. _(*does some intense Vulcan meditation*)_
- Reprompted me for access for dev and created a user token at https://myaccount.google.com/security#connectedapps _(**EDIT**: Seems to be dev only... pro didn't do this)_
- Strategy name is still stored as `google` which is as it should be
- "key" is different for dev and pro now ... this might get ugly when on production with their prompting of "Somethings different about you..." assuming that happens to everyone.
- Initially confusing repo name vs. package name for the uninitated dev side ... but this won't be the first time a package has a different name than the repo. _(**EDIT**: Seems to be dev only... pro didn't do this... possibly detection of IP change)_

@sizzlemctwizzle 
We'll need to check the migration routine you put in and make sure it doesn't collide somewhere... e.g. remove that code is my guess from #484 _(PR #564)_

Ref:
- https://github.com/OpenUserJs/OpenUserJS.org/issues/83#issuecomment-199075548
- https://github.com/jaredhanson/passport-google-oauth#passport-google-oauth

Martii:
This issue is on hold for the moment. e.g. do not resolve it.

PR #1527, opened 2018-10-08
Swap to Goo relative profile scope

Martii:
* Setting goo auth to read only for a while until this is clarified at passport strategy

NOTE:
* Untested on pro but seems to work on dev

Applies to #1526

### DIFF
--- a/controllers/auth.js
+++ b/controllers/auth.js
@@ -86,7 +86,7 @@ exports.auth = function (aReq, aRes, aNext) {
     }
 
     if (strategy === 'google') {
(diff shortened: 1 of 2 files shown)


Martii:
Tested on pro... pass... however dev showed me goo authorization card/panel to access account whereas pro did not.

Martii:
One other note from historical... https://github.com/OpenUserJS/OpenUserJS.org/issues/889#issuecomment-383551771 ... back then when I tested this `profile` change with a new account it tripped the server... not sure what it does now but when the dust settles will reexamine... hence why the passport google is read only atm.

PR #1529, opened 2018-10-14
Fix to non-alias `xpath` of `xquery` base for *highlight.js* dep

Martii:
* Add in some dev console messages to assist

Post #1528 #1438 ; Applies to #430 ; See also highlightjs/highlight.js#1868


Martii:
Local pro ex.

``` console
Dependency automatic named highlighting failed with:
TypeError: Cannot read property 'disableAutodetect' of undefined
```

Dev ex. may contain further stack trace.

PR #1532, opened 2018-10-28
Rework cert detection a bit

Martii:
* This is a WIP... so there might be an error thrown when running on actual pro *(hopefully not)*... working on it now, so possible downtime expectation perhaps. Works on local pro so far.
* Check for backup alternate keys
* Ensure keys are in UTF-8

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,10 +1,5 @@
 'use strict';
 
-// Define some pseudo module globals
-var isPro = require('./libs/debug').isPro;
-var isDev = require('./libs/debug').isDev;
-var isDbg = require('./libs/debug').isDbg;
-
 // Stamp a message for stdout...
 console.log('Starting application...');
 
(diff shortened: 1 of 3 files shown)


Martii:
Retested this with last years certs and it produces a general failure on actual expired certs:

``` console
$ node app.js
Starting application...
_tls_common.js:134
      c.context.setKey(key, passphrase);
                ^

Error: error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch
    at Object.createSecureContext (_tls_common.js:134:17)
    at Server (_tls_wrap.js:868:27)
    at new Server (https.js:62:14)
    at Object.createServer (https.js:84:10)
    at Object.<anonymous> (~/OpenUserJS.org1/app.js:303:24)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:741:12)
    at startup (internal/bootstrap/node.js:285:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:739:3
```

e.g. if they expire it will currently, constantly, trip the server with this message with process manager *(pm)*. May need a `try...catch` in there ... however pm won't know any different... renamed pub to something else and it did run the server however not reachable through the browser. e.g. didn't trip the server but not accessible.

PR #1537, opened 2018-11-07
Add capability of donations

Martii:
* Site donations
* Adopt `@contributionURL` ... must be secure web address... expandable to filtering if needed.
* Remove dismiss ability on `@supportURL`
* Option'ify the `isFQUrl` since it's getting bigger on args
* Add current to add script page docs
* Users will need to update their script to show this... may run through the DB and recollect this but that's bandwidth intensive... so don't count on it atm.
* Needs potential followup for possible `referrerpolicy` change for Author donation URLs.
* Slight wording change under etiquette since we are accepting donations now.
NOTES:
* `@contributionAmount` is kind of pointless since it can be anything... could be a text label but if utilized other than a label, currency conversion is an issue... for the time being leaving this at the orphanage.
* Donate for Authors is located on the Issues page of script Authors... one should see what's sort of praise, bugs, etc. before donating.
* Tried to make this as unobtrusive as possible without sacrificing emphasis.

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -1529,7 +1529,7 @@ exports.storeScript = function (aUser, aMeta, aBuf, aUpdate, aCallback) {
 
       icon = findMeta(aMeta, 'UserScript.icon.0.value');
       if (icon) {
-        if (!isFQUrl(icon, false, true)) {
+        if (!isFQUrl(icon, { canDataImg: true })) {
 
           // Not a web url... reject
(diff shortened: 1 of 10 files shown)


Martii:
Deployed.... just barely. GH is having issues at the moment. 

Added OUJS payment processor id link for PP to RFC 2606§3 - Hello, World!

Martii:
One other note... since common.css was changed do a hard refresh of your browser or clear your cache to show those changes.

PR #1552, opened 2018-12-19
Increase `poolSize` depending on multiplier

Martii:
* This didn't seem to help in a direct test on production but doing since it's the logical thing to do with our current process manager.

NOTE(S):
* We don't currently have clustering management in the project itself but may at some point in the future so this could eventually use some improvement. Trying to keep this simple at start for everyone.

Applies to #1548


Martii:
NOTE: Worked for a little bit but then resumed lagging.

PR #1576, opened 2019-01-27
Isolated dep update

Martii:
* Untested as I don't do FB
* Commit history doesn't seem to affect semver major. Needs testing with someone who has it.


Martii:
Full auth period detected with FB... seems to be working. :smile_cat:

Issue #306, opened 2014-08-20
Case insensitive Sort Order


TimidScript:
Sort order should be case sensitive and would be nice if it reflects an order similar to windows listing or Greasy Fork. Brackets and characters like `* ! ; @`  should come before alphanumeric characters.
- Ability to Sort order Groups by size. **(MARTII EDIT)** ... already present **(/MARTII EDIT)**

Martii:
See also:
* http://stackoverflow.com/questions/10728043/mongo-query-with-regex-in-node-js-operating-on-a-variable/10728069

Martii:
Followup:

> available in MongoDB 3.3.11 via the new collation feature.

Martii:
> P.S. Scheduled a server backup.

[completed](https://openuserjs.org/announcements/Server_Maintenance?p=2#comment-15b75647c3c).

Martii:
We've been able to do this for a while on the MongoDB versions even on mLab ... so applied level 3.

Ref:
* https://docs.mongodb.com/manual/reference/collation/

Notes:
* Needs testing to see if this is complete and as efficient as possible with the PR merged... then close. TimidScript is no longer with us as per policy.
* Marking as a bug since I think it is after all of this time pondering and waiting.

PR #1578, opened 2019-02-02
Collation on `find` for applicable models (WIP)

Martii:
NOTES:
* Currently using level 3 for a test
* This is highly localized to `en` since that is the default language of OUJS. Needs mitigation when multiple languages are implemented. e.g. there will be languages that aren't sorted. `simple` doesn't appear to do anything useful in testing.
* May need check to ensure that whole list isn't being pulled from DB. Callback not present at these stages so guessing it isn't doing this but could be misunderstanding this. Cc: @sizzlemctwizzle @jerone @Zren *(any thoughts?)*

Applies to #306

### DIFF
--- a/controllers/discussion.js
+++ b/controllers/discussion.js
@@ -131,7 +131,7 @@ exports.categoryListPage = function (aReq, aRes, aNext) {
   options.multipleCategories = true;
 
   // discussionListQuery
-  discussionListQuery = Discussion.find();
+  discussionListQuery = Discussion.find().collation({ locale: 'en', strength: 3 });
 
   // discussionListQuery: remove issues
   discussionListQuery.and({ issue: { $ne: true } });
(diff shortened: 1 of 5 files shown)


Martii:
Test samples:
* http://localhost:8080/users/Marti/scripts?orderBy=name&orderDir=asc vs http://localhost:8080/users/Marti/scripts?orderBy=name&orderDir=desc **PASS**
* http://localhost:8080/?orderBy=name&orderDir=asc&p=1 vs http://localhost:8080/?orderBy=name&orderDir=asc&p=2 **FAIL??** *(look for `hello, world` vs `Hello, world`... has `RFC 2606§3 - Test relative require OUJS url` in between these... still thinking on why this is happening... okay on individual script list but main list is moving it somehow)*

Martii:
Going to go ahead and merge test this on pro... if there is an issue primary issue is still open at #306 . Seems to work on my end but still could use additional feedback.

Martii:
Deployed.

PR #1585, opened 2019-02-10
Fix bug with `votes` count

Martii:
* Script up to down or down to up improperly counts script votes and messes up good/bad bar. Shouldn't affect Rating in Script or Group but will recheck a few.
* At the same time move to MVC *(Currently isolated to Script e.g. parallel change)*... Fixes TODO in routes.js
* Some styleguide.md conformance
* Simplify some views. e.g. turn some things off when they aren't able to be used. Save b/w and possibly improve understanding for most
* Still would prefer returning `err` *(aErr)* with libs but that's probably a separate PR since there are still inconsistent return callback parms
* Some filling in for #1548
* Some additions for #430. e.g. we want to know if a DB action fails currently

NOTES:
* Structure mirrored and altered from `flagsLib`
Applies to #262 and Active Maintainer override (Sorry sizzle but needed to be done so it doesn't mess up other areas down the line)
* Will run through scripts and mitigate any script `votes` counts.

### DIFF
--- a/controllers/script.js
+++ b/controllers/script.js
@@ -16,7 +16,6 @@ var SPDX = require('spdx-license-ids');
 var Discussion = require('../models/discussion').Discussion;
 var Group = require('../models/group').Group;
 var Script = require('../models/script').Script;
-var Vote = require('../models/vote').Vote;
 
 //--- Controller inclusions
 var scriptStorage = require('./scriptStorage');
@@ -29,6 +28,7 @@ var getFlaggedListForContent = require('./flag').getFlaggedListForContent;
 
 var isSameOrigin = require('../libs/helpers').isSameOrigin;
 
+var voteLib = require('../libs/vote');
 var flagLib = require('../libs/flag');
 var removeLib = require('../libs/remove');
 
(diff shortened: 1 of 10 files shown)


controllers/script.js
```diff
@@ -506,125 +488,3 @@ exports.edit = function (aReq, aRes, aNext) {
...
-              aScript.votes = votes <= 0 ? 0 : votes - 1;
-              if (oldVote) {
-                flags = 1;
-              }
-              saveScript();
-            });
-            return;
-          } else if (aVoteModel.vote !== vote) {
-            aVoteModel.vote = vote;
-            aScript.rating += vote ? 2 : -2;
```
Martii: The bug of missing inverse the `votes` count.

libs/vote.js
```diff
@@ -5,4 +5,252 @@ var isPro = require('../libs/debug').isPro;
...
+      aScript.votes = 0;
+    }
+
+    votes = aScript.votes;
+
+    if (aVote) {
+      if (aVote.vote !== aCasting) {
+        aVote.vote = aCasting;
+        aScript.rating += (aCasting ? 2 : -2);
+        aScript.votes = aScript.votes + (aCasting ? 2 : -2);
```
Martii: The bug fix of missing inverse the `votes` count.

PR #1603, opened 2019-04-24
Migrate to *node*@12.0.0

Martii:
* Adjust default cipher order to match. Support for TLS 1.3 is added in this version.
* Remove restriction on TLSv1.0 since it's presumably disabled by default.

NOTE:
* This is about a year early from the March-April'ish 2020 recommended deadline.


Martii:
Btw this kill *kerberos* compilation... but that is purely optional and I'm sure they'll release another semver at some point.

PR #1615, opened 2019-05-30
Always show the comment reply form

Martii:
* Allows some browser back button to work in certain browsers due to logoff condition. Edge use case.

NOTE:
* This is a **work-around** for those browsers not implementing back-browser action in a consistent non-destructive manner. Currently present in Chromium back-ends but not Mozilla based or Safari based. Nice inconsistency google! Perhaps this is a bug to add to their collection. Probable refetch instead of cache pull *(and listen)* in those affected browsers which can be another imposed DoS vector.
* No known current way, w/o additional DoS vector added, to address for newTopic logoff condition e.g. out of luck atm.

Post #1204

Refs:
* https://openuserjs.org/scripts/AltoRetrato/IMDb_My_Movies_enhancer/issues/does_it_still_work!#comment-16af97e508e *(subtle report not yet done on development)*
* https://stackoverflow.com/questions/39920118/chrome-forces-page-refresh-on-window-history-back *(unanswered and long time issue)*
* https://github.com/OpenUserJS/OpenUserJS.org/issues/1548#issuecomment-474667560  *(about:flags note ... perhaps there is another one to disable bad browser acting)*

### DIFF
--- a/views/includes/commentForm.html
+++ b/views/includes/commentForm.html
@@ -26,7 +26,7 @@
               <div class="submit-panel btn-toolbar">
                 <a href="/about/Frequently-Asked-Questions"><i class="fa fa-question-circle" title="FAQ"></i></a>
                 <a href="https://guides.github.com/features/mastering-markdown/"><i class="octicon octicon-markdown" title="GitHub Flavor Markdown compatible"></i></a>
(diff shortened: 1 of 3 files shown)


Martii:
Works on dev but not on pro... hmmm.

PR #1620, opened 2019-06-12
A dep update/change

Martii:
* Force update *s3rver* migration
* This is a breaking change for dev. Project version bump. All dev sources previously there must be reimported/rewritten/reuploaded *(again, from when we left *fakes3* in #1520 )*. e.g. this is not forward or backward compatible.
* Bucket configuration and CORS config doesn't appear to be needed on *s3rver*

NOTES:
* Untested on pro but should probably work since it's been working this way. Dev Source Code tab pulls and no bucket errs. Will test momentarily on pro to ensure it is 100%.
* Documentation is still dispersed for this dependency and partially pulled from example folder
* Seems to be more inline with pro options for *aws-sdk*

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -128,12 +128,11 @@ if (isPro) {
 } else {
   DEV_AWS_URL = process.env.DEV_AWS_URL || 'http://localhost:10001';
   AWS.config.update({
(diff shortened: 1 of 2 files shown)


Martii:
Existing source still available... PASS

PR #1631, opened 2019-06-30
Abstract some limits

Martii:
* Put in env var check for limiting problem query

Post #1629 and applies to #1548


Martii:
Btw I've tracked a bunch of IPs that will be getting abuse reports for this at minimum. While I appreciate the test data it's getting old that some don't know how to use the search field. Think `github汉化`. Try using a space in between ASCII and non-ASCII. It does make a difference.

PR #1646, opened 2019-08-07
Disable `Bad escaping of EOL. Use option multistr if needed.`

Martii:
* Multi-line strings like this, especially in the original JavaScript spec, are still valid and supported although Template Literals (Strings) are a bit easier to read. e.g. don't nag with this.

Post #1641

### DIFF
--- a/views/includes/scripts/scriptEditor.html
+++ b/views/includes/scripts/scriptEditor.html
@@ -362,7 +362,8 @@
                 'eqnull': true,
                 'sub': true,
                 '-I003': true, // "ES5 option is now set per default"
-                '-W014': true  // "Bad line breaking before '{a}'."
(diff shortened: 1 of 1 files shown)


Martii:
Oy!

Ref:
* [ajaxorg/ace/blob/`c50e196`/lib/ace/mode/javascript_worker.js#L77-L98](https://github.com/ajaxorg/ace/blob/c50e196269f7cc84edd8197676a1c22155d2e6a8/lib/ace/mode/javascript_worker.js#L77-L98) ... defaults that get overwritten. Will revisit later to add some of these in as well. Knew there was this but took a really long time to find them. Sure would be nice to have a "getOptions" to see what they are by default to modify.

PR #1668, opened 2019-08-25
More icons with tooltips for the UI

Martii:
* Squeeze in a little more real estate for more the more important things. This is as small as I dare make it and still be readable which in turn adds some vertical white-space occasionally.
* Fixing a Y2K+100 bug ;)

NOTES:
* Tested on desktops only... will do more testing after merge. Hopefully everything goes well.
* Additional squeezing yet to come.


Martii:
Btw this should be making room for labeling issues at some point. As I mentioned already "iconic" is about as small as it's going to get without compromising readability for the time being.

Issue #1705, opened 2020-02-05
GitHub API Deprecations (repoManager and GH import pending failure)


Martii:
Got this in the email today *(which is a little weird since I don't have OUJS GH access... but none-the-less)*:

> Hello there!
>
> On February 4th, 2020 at 13:20 (UTC) your application (OpenUserJS) used its `client_id` and `client_secret` (with the User-Agent NodeJS HTTP Client) as part of a set of query parameters to access an endpoint through the GitHub API:
>
> https://api.github.com/user/_useridclipped_
>
> Please use Basic Authentication instead as using OAuth credentials in query parameters has been deprecated.
>
> Depending on your API usage, we'll be sending you this email reminder at most once every 3 days.
> Just one URL that was accessed with a User-Agent combination will be listed in the email reminder, not all.
>
> Visit https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters for more information.
>
> Thanks,
> The GitHub Team

Target code point is probably [https://github.com/OpenUserJS/OpenUserJS.org/blob/`1261d16`/libs/repoManager.js#L67](https://github.com/OpenUserJS/OpenUserJS.org/blob/1261d161171f0558aa27d4de97f4a9a6c4fc9a15/libs/repoManager.js#L67).

I don't have much time to work on this but it seems that GH importing will fail eventually when it is EOL'd after deprecation. It is open for assignment atm.

See also:
* https://developer.github.com/v3/auth/#via-oauth-and-personal-access-tokens
* https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/

@sizzlemctwizzle 
Seems like you will need to set up something with OAuth if I'm skimming this correctly.

For everyone else when/if it breaks, it breaks... the work around should be to paste your source into the browser from github raw url. It would be nice to have this OUJS feature continued as a lot of people do use it. As far as my past changes to the webhook this theoretically shouldn't change anything there. i.e. if your OUJS account has GH authentication then the webhook should still process push notices.

Cc: @Zren

Martii:
Hmmm. Thinking out loud for a moment:

* https://developer.github.com/v3/#authentication

Re:

> OAuth2 token (sent in a header)
>
> curl -H "Authorization: token OAUTH-TOKEN" https://api.github.com
>
> Note: GitHub recommends sending OAuth tokens using the Authorization header.
>
[...]

PR #1729, opened 2020-08-09
Use Basic auth for GitHub (#1705)

joeytwiddle:
I don't know if this will work, since I don't have a full test setup.

However this is how we did Basic authentication in our app's fetch requests. So fingers crossed it will work for GitHub too!

Addresses issue #1705

### DIFF
--- a/libs/repoManager.js
+++ b/libs/repoManager.js
@@ -28,7 +28,7 @@ Strategy.findOne({ name: 'github' }, function (aErr, aStrat) {
 });
 
 // Requests a GitHub url and returns the chunks as buffers
-function fetchRaw(aHost, aPath, aCallback) {
+function fetchRaw(aHost, aPath, aCallback, aOptions) {
   var options = {
     hostname: aHost,
     port: 443,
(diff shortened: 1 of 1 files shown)


Martii:
@joeytwiddle 

Had some more dedicated time to think about this... seems syntactically correct although I'm wondering where you go the `Basic` authorization header syntax from? I see a modified header at https://developer.github.com/v3/auth/#using-the-oauth-authorizations-api-with-two-factor-authentication but nothing regarding actual basic authorization headers. Thanks for citing please.

We'll test it for a while directly on pro and if everything goes well *(you'll have to tell me and I'll watch the logs/possible restarts)* will merge it unless @sizzlemctwizzle or other steps in with a concern.

Martii:
This appears to be working ... specifically not emailing us... going to merge since there is no other response. Doesn't seem to hurt things in the few webhook tests I've done and I'm sure someone would have barked by now. Would be nice to know the citation on where the `Basic` header came from... still can't confirm it.

joeytwiddle:
Some documentation:

- Describes the Basic scheme simply, but well enough to implement it for clients:

    https://en.wikipedia.org/wiki/Basic_access_authentication

- Links to RFCs for Basic, Bearer, Digest and other authentication schemes:

    https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication

I'm sure there are plenty more docs and tutorials out there. But not sure which is the "best one".

----

The scheme we used in our own projects (and I think the most commonly used) was to send "Basic <username+pass>" to request a token, and thereafter use "Bearer <token>" for all subsequent requests.

This takes a little more code because of the need to persist the token between requests, and the possibility that it might one day expire.

Issue #1730, opened 2020-08-09
WebHooks aren't being processed?


joeytwiddle:
I often push updates to my userscripts, but it seems **some of** the new versions haven't been loaded from github for some time, whilst others have: https://openuserjs.org/users/joeytwiddle/scripts

For example, this one has changed a few times (see its History): https://github.com/joeytwiddle/code/blob/master/other/gm_scripts/Related_Links_Pager/Related_Links_Pager.user.js

**Do you have some debugging information about which part of the hooks process has a problem?**

The script mentioned above is 50KB, well within the 1MB limit specific in `settings.json` (at least in the dev settings).

I see the hooks process starts at `exports.webhook` in `scriptStorage.js` and goes on to call `repoManager.prototype.loadScripts` in `repoManager.js`.

I would assume the `fetchRaw()` is working as it is, but if you want me to add authentication to that, I could do that.

----

I just did a push to GitHub at `new Date(1596957303135)` so perhaps you might see something logged around that time.
https://openuserjs.org/scripts/joeytwiddle/Wikipedia_Inline_Article_Viewer
https://github.com/joeytwiddle/code/blob/master/other/gm_scripts/wikipedia_inline_article/wikipedia_inline_article.user.js

Martii:
@joeytwiddle 

> Do you have some debugging information about which part of the hooks process has a problem?

We do not. We return the http status codes that shows up in your webhook logs on GH only. Browsing through the `console` messages in our source for errs I see nothing related to your account on OUJS. May have an issue with *mongoose*/*mongodb* dep though... updating those shortly.

> I would assume the fetchRaw() is working as it is, but if you want me to add authentication to that, I could do that.

Well as you know it probably isn't. I'll look at the PR in a bit but I'll probably have to try that VPS side directly since dev environment doesn't even process the webhook. If @sizzlemctwizzle would be available to review #1705 that would be appreciated.

> **some of** the new versions haven't been loaded from github for some time

I see a date updated difference on one of those. Although I just did my usual "Hello, World" test push and it sync'd well. You could be running into GH throttling with #1705 but unsure atm.

joeytwiddle:
Recently on my Syncs page I am getting a lot of "Updating but no script found."

I think some years ago GM for Firefox might have lowercased all my script names, and I went along with it.

I can rename the file, but this is a bit of a hassle (and breaks some of my own processes).

Would it be possible to make the `findDeadorAlive()` query case-insensitive?

[...]

PR #1735, opened 2020-08-17
Change status code for script not found

Martii:
Instances:
1. Author has webhook enabled and GH sends update notice before imported on OUJS.
2. Author deleted OUJS script and still receiving webhook notice.

There's still the extremely rare chance that this could be a 500, if *mongoose*,*mongodb*, and S3 has a catastrophic connection error.

Since the bulk of this is usually client error denoting as such in case GH is paying attention and possibly doing something on their end by killing further notices. If true then this part "could be" related to #1730

Post #1731


Martii:
Deployed for further testing.

Issue #349, opened 2014-09-16
Model `published` and/or `updated` *(similar to `created`)* field


Zren:
With MongoDB. `_id: { type: ObjectId }` (which is the default id type) is made up of a timestamp of when the object was created (plus a few other things to make it unique across servers when you distribute the db). We can use it right now to fetch the creation date of scripts and other models.

It's currently possible to sort by creation date like so:
https://openuserjs.org/?orderBy=_id&orderDir=asc

To display it, we can use `_id.getTimestamp()`, but I think making a virtual _(**OWNER EDITED**: static stored)_ property `BaseSchema.created` would be best.

Martii:
@Zren 

Sorry didn't see this ticket and probably didn't understand it at the time it was posted...anywhooo...

Which other models did you want this added to?

I already added it to `Script` and `User` in #485... we should continue to use `_since` as the virtual property name... and did you want to do anything different on the hard-coded dates? I may have to email sizzle to see if his commenting routine *(`Discussion` and `Comment`)* will be blocking since he is assigned to that as well as `Vote` ... and the virtual will currently need to check the existing creation date property first and return that instead.

Zren:
Every single model can benifit from a `createdAt` and `updatedAt` column. You can't query over virtual functions (and we shouldn't really be querying on ObjectId since it includes other stuff). You can sort to find a new discussions, new votes, new groups, new ____.

> did you want to do anything different on the hard-coded dates?

Eh? Are you talking about `ObjectId.getTimestamp()`? All I would use it for is populating the createdAt column if it was added.

Martii:
> You can't query over virtual functions

Please cite reference for this from *mongoose* package API docs. I'll try a query at some point in the near future to confirm. Thanks.

> All I would use it for is populating the createdAt column if it was added.

Depending on the citation and testing... this is doable... although if createAt isn't necessary we should avoid hard-coding it into the schema.

Zren:
Virtual functions aren't serialized into the db, nor are their calculated values, ergo you can't query over them.

http://mongoosejs.com/docs/guide.html#virtuals

Martii:
> http://mongoosejs.com/docs/guide.html#virtuals

"Only non-virtual properties work as part of queries and for field selection." 

[...]

PR #1748, opened 2020-09-04
Update copy of dev DB

Martii:
Applies to #349


Martii:
mLabs dev DB not migrated yet... will have to do this in *mongoose* since MongoDB versions differ... need sleep first.

Issue #1556, opened 2018-12-20
Consider ES6+ migration


Martii:
The project is currently ES5.1 and it would be helpful to migrate so we can use some more features.

1. Some of the CONTRIBUTING.md guidelines would need to be changed... especially regarding "hoisting" since the `let` syntax would change if it was coded hoisted. We are using some `let` syntax already for speed/memory considerations in key code points.
2. I am totally **not** into the arrow operator and it actually makes code more difficult to read and improves laziness. :-1:
3. With the recent issue at #1548 it may be prudent to dump *async* at some point since `async` and `await` exist. There are a few other technologies too but `Promises` are also a :-1: from me.
4. I do like callbacks but when combined with our current site issue and *async* it seems to be error prone. Perhaps a combination of these to be allowed in CONTRIBUTING.md

While I'm at #1548 I'll do some pro tests to see how well it handles that soon after the holidays.

Open to discussion here.

Martii:
Note:
* #1709 deprecation of *request* may force some `async`/ `await` usage i.e. promises

PR #1793, opened 2021-04-13
Additional dep

Martii:
* Partially migrate to newer dep for GH API
* Adding some error protection, **but not all**, to the GH API deprecation of QSP's. This is probably temporary until a broader fix is implemented. GH importing may have to be disabled after May 5th, 2021... webhooks might be affected if the old (or new) Promise rejection fails. i.e. no more webhooks... but we'll see after the 5th of next month if not addressed.

Applies to #1705

NOTE(S):
* Due to the nature of the dependency with Promises it is currently contrary to the STYLEGUIDE.md See #1556 as is #1729 with ES6+ syntax.
* Doing what I can to avert this chaos but the Code is spread out all over the place.


Martii:
Might of hit the "brown out" period yesterday/today with test results... so embedded code comments might be invalidated.

PR #1798, opened 2021-04-16
Enable usage of the import kill switch

Martii:
* Disables *github* dep authorization which in turn disable importing. When activated use paste to site via Write Script Online or file upload via Upload Script until the dep can be successfully migrated. If authorization is enabled and there's a dep issue a 500 Server error could occur if GH changes something.
* More bookmarks... can't quite tell if it's Fx but the bookmarks are shooting low on auto browser scroll... will try another browser at a future date.
* Some verbiage clarity.

Applies to #1705

### DIFF
--- a/libs/githubClient.js
+++ b/libs/githubClient.js
@@ -31,7 +31,7 @@ Strategy.findOne({ name: 'github' }, async function (aErr, aStrat) {
   if (aErr)
     console.error(aErr);
 
-  if (aStrat) {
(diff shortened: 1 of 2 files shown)


Martii:
> can't quite tell if it's Fx but the bookmarks are shooting low on auto browser scroll... will try another browser at a future date.

* Pale Moon 29.1.1 seems to be working correctly with autoscroll

PR #1801, opened 2021-04-17
Enable debugging of *github* dep

Martii:
* Also at some point *node* changed the way to detect if debugging was in process. Updated original comment at https://github.com/OpenUserJS/OpenUserJS.org/pull/429#issuecomment-63425552 Been a while since I've been in this mode for deep debug examination.
* Fix missing custom headers for UA. This is optional but didn't know this until now that it had that option for the *github* dep. Post #1753

Applies to #1705

Ref(s):
* https://nodejs.org/api/inspector.html#inspector_inspector_url
* https://nodejs.org/api/debugger.html

### DIFF
--- a/libs/debug.js
+++ b/libs/debug.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var inspector = require('inspector');
 var fs = require('fs');
 var os = require('os');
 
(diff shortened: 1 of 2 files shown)


Martii:
YAY... sorta... the email notice included the custom UA that was put in here via #1753

Issue #1825, opened 2021-07-18
Migrate from MongoDB 4.x to 5.x


Martii:
1. Check with backend to make sure VPS is supported *(hopefully)*
2. Take OUJS offline
3. Compact DB
4. Backup existing pro DB *(ex. `mongodump --gzip --db openuserjs_devel --archive=./dev/devDBdirty.gz` but productions)*
5. Purge 4.x
6. Target new sources
7. Install 5.x *(may need db tools from .deb package... local installation was a bit weird with `apt`)*
8. Import pro DB *(ex. `mongorestore --gzip --db openuserjs_devel --archive=./dev/devDBdirty.gz` but productions)*
9. Put OUJS online *(possible in read only mode just for initial checking)*

NOTE(S)
* `mongo` command is deprecated for `mongosh`
* Time it took locally was about 1 hour ... so approximately 1 hour downtime. Schedule at some point with notice on OUJS.

Martii:
This is starting to get important. `mongo` is starting to show bitrot as an updated element didn't update but it said it did. Need `mongosh` soon as it is more actively developed. Blehh.... more work for the new year as usual.

See also:
* https://github.com/OpenUserJS/OpenUserJS.org/pull/1899#issuecomment-1000253046

Martii:
Interesting:

* https://docs.mongodb.com/manual/core/index-multikey/

MongoDB 4.x seems to allow one only on production however local dev here says the index was created when I did **2**. Since pro is a slightly different platform may not work with 2... but now there is an option. Been waiting for this feature for several years now. We can query multiple "parallel arrays" for years however we can't index them is what they are saying in the docs... paraphrased of course.

Martii:
Misc critical error encountered:

``` console
MongoDB Script save critical error
iFelix18/Monkey_Utils.js
{
 "version": 0,
 "modifiedPaths": [
  "meta",
  "uses",
  "hash",
  "updated"
 ]
}
0bc2e3b4-803a-11ec-868a-26da9250a6e2 statusError: {
 "message": "Database write error",
 "code": 502
}
    at ~/OpenUserJS.org1/controllers/scriptStorage.js:2178:27
    at ~/OpenUserJS.org1/node_modules/mongoose/lib/model.js:5074:18
    at processTicksAndRejections (node:internal/process/task_queues:78:11) {
  status: { message: 'Database write error', code: 502 },
  code: undefined
}
```

[...]

PR #1834, opened 2021-08-18
Some dep updates

Martii:
Please read their individual commit summaries.


Martii:
Some notes on future migration:

* *marked* is severely broken on tables so skipping atm
* *remark* moved to ESM instead of CommonJS and is loosely related to #1556 so skipping atm until "divine intervention" steps in.
* *strip-markdown* also moved to ESM instead of CommonJS ... see prior bullet point
* *spdx-license-ids* diff log is getting rather large and holding off migration since, to my knowledge, there haven't been any addition FOSS licenses... will eventually address this with more SPDX blocks when available time.
* *mongodb* is potentially breaking as mentioned at #1825

PR #1835, opened 2021-09-02
*node* bump

Martii:
* This finally warns using *npm* under `engines` with *node*
* Also startup with strict instead of lax. Has been in place for about a month as a test for login issues. Still `lax` is needed for OAuth/etc. unfortunately.
* Delete op, sanitize, markdown, moderation flagging, overall UI... retested
* Move git repo linkage in package.json to https. Seems like https://github.blog/2021-09-01-improving-git-protocol-security-github/#how-do-i-prepare with "anonymous" git might be affecting this from what I gather. See also https://docs.npmjs.com/cli/v7/configuring-npm/package-json#git-urls-as-dependencies . Works on current dev stack... haven't ever tried this on pro... so \*crosses things\* Indirectly applies to #1705 . After March 15, 2022 can retry `git:` but preventative atm.

### DIFF
--- a/app.js
+++ b/app.js
@@ -422,7 +422,7 @@ app.use(session({
   cookie: {
     maxAge: 5 * 60 * 1000, // minutes in ms NOTE: Expanded after successful auth
     secure: (isSecured ? true : false),
-    sameSite: 'lax' // NOTE: Current auth necessity
+    sameSite: 'strict'
   },
   rolling: true,
   secret: sessionSecret,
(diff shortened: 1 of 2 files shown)


Martii:
Further Ref: Cc: @sizzlemctwizzle 

* https://github.blog/2021-09-23-announcing-npms-new-access-token-format/
* https://www.npmjs.com/settings/{username}/tokens/

I'm not sure what it going to happen when there isn't a token.

PR #1840, opened 2021-10-16
Accept Github webhook requests for the Github default branch

### LINKED ISSUE #1781: Support github sync main branch
Github adopted main as default branch name. But it seems to cause 403 with the openuserjs github webhook sync feature.
How about supporting it?

phw:
This allows the webhook to work independent of the name of the default branch on Github, which used to be master, but is now main. The branch can also be renamed by the repository owner to anything they like, e.g. I have seen teams working with `default` or `develop`. 

This PR ensures the webhook works with whatever the main branch is called. Github reports it in the webhook payload as follows (this shows only the relevant details for branch handling, every other detail is omitted):

```json
{
  "ref": "refs/heads/main",
  "repository": {
    "default_branch": "main"
  }
}
```

Fixes #1781

### DIFF
--- a/controllers/scriptStorage.js
+++ b/controllers/scriptStorage.js
@@ -2223,8 +2223,9 @@ exports.webhook = function (aReq, aRes) {
 
   //
 
-  // Only accept commits from the `master` branch
-  if (payload.ref !== 'refs/heads/master') {
+  // Only accept commits from the default branch
+  defaultBranch = `ref/heads/${payload.repository.default_branch}`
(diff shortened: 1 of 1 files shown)


phw:
> It's in other issue discussions. It is what it is.

Ok, I still don't see why it shouldn't be possible to support more than one name for the primary branch or why migration to main as default would be a one way road. But if you say it has been discussed somewhere in the past let's leave it at that.

Hopefully someone will figure out a way to support at least main in the future. Thanks for the feedback.

PR #1850, opened 2021-11-20
Modify placeholder for new scripts

Martii:
* Also move the options for this out regardless of state/mode... going to try something in a while to further assist there.

Post #1847

### DIFF
--- a/controllers/user.js
+++ b/controllers/user.js
@@ -2409,6 +2409,11 @@ exports.editScript = function (aReq, aRes, aNext) {
     });
   });
 
+  // Lockdown
+  options.lockdown = {};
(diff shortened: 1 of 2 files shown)


Martii:
> going to try something in a while to further assist there.

This may be stalled as I just discovered a "change" between SeaMonkey and Fx/Chromium i.e. Fx/Chromium something that should normally work doesn't *(could be a jQuery issue as well)*. 

Still twiddling... thanks for your patience.

Martii:
Confirmed working assistance draft in SeaMonkey 2.53.8.1 *(will update later to latest and retest)* and failure in retest with Fx 94.0.1

UGHH!

PR #1899, opened 2021-12-23
Collect update time to User

Martii:
* Going to need these in a next phase

Applies to #349

NOTE:
* Will migrate the DB and bump the version when it happens but not right away... since it's rather complex.


Martii:
Lost ~2 hours worth of DB contents due to a corruption error.... Apologies... at least we had a backup.

PR #1913, opened 2021-12-30
Ease some caps and limits for privileged authed

Martii:
* Mod's need more access *(less cap limiting)* and of course Admin+.
* Rate limiting is in effect for Mods in case their device or network is compromised. May consider a "pardon" like with scoring when that time comes.

Applies to #944

NOTE:
* This is a WIP. Ton of things to test and (re)consider. Patience is a virtue. ;) :)

### DIFF
--- a/routes.js
+++ b/routes.js
@@ -51,6 +51,14 @@ var installCapLimiter = rateLimit({
   handler: function (aReq, aRes, aNext, aOptions) {
     aRes.header('Retry-After', waitInstallCapMin * 60 + 60);
     aRes.status(429).send();
+  },
+  skip: function (aReq, aRes) {
+    var authedUser = aReq.session.user;
+
+    if (authedUser && authedUser.isMod) {
+      this.store.resetKey(this.keyGenerator);
+      return true;
+    }
   }
 });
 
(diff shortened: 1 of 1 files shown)


Martii:
@jesus2099

What also does this commit summary say to you?

This is also in reference to the assumption you put on production.

jesus2099:
Is it also about the captcha problem I got when trying to edit my profile a second time? https://openuserjs.org/discuss/Edit_Profile_Captcha_is_broken_429_Too_Many_Requests

Martii:
@jesus2099 

> Is it also about the captcha problem I got when trying to edit my profile a second time?

Yes. Look at the actual commit message fourth here where it says:

> "This is a WIP. Ton of things to test and (re)consider. Patience is a virtue. ;) :)"

In short, outside of the teaching exercise for you, things are changing to combat some bad actors trying to take the site down. Things will definitely be in flux for some time and I only have so much energy and time at the moment to implement it all. We are also in lockdown which is a feature that has been around since 2014 to keep the site up. Unless you have a reason to take the site down these caps and limitations will stay in effect until the threat is over. I hope you understand a little more about this. I always do my best to document the commit history in as plain, and clear, as I can get put it at the time. :)

PR #1931, opened 2022-01-10
Rate limit statics

Martii:
* Might be interesting to see the effects on proxies... but as I alluded to elsewhere a bad apple makes the whole bunch bad imho.

Post #944


Martii:
Impressed... first time I've seen a server from Qatar.

PR #1964, opened 2022-06-24
Pull1963refix

Martii:
See individual commits


Martii:
Complete failure on production for previews.. .however partially working on development.... UGGH!

Martii:
Forgot *bootstrap-markdown* is tied to this... need to see if `parse` is supported... could be failure on pro however doesn't explain why it's partially working in dev. Major UGHH!

PR #1991, opened 2023-02-26
FAQ: stop recommending "jquery-latest.js"

rybak:
Fix for an outdated recommendation in the FAQ. Details are in the commit message.


Martii:
Thanks for the PR... however there is some history to this. One we tried to get jQuery to update that virtual link to the latest and they had a very large cow at the mere mention.

Additionally we don't want to advocate one version over the other and the way the link should have been done with that framework is to move the latest to the real latest... however that's not going to happen and we're not going to advocate a static version either.

Do appreciate the effort but this is vetoed.

PR #2056, opened 2025-03-02
Enable logographic detection

Martii:
* Trim up some of the data a bit who don't know how to use SEO properly. i.e. possible spamming/poisoning
* Bug fix using incorrect identifier.

Post #2048

NOTE(s):
* I tried for HOURS to get these settings into settings.json *(hence why the require is left in one file)* however RegExp just wouldn't have it no matter what I tried. Existing scripts worked, new scripts did not. If anyone has an idea PRs are welcome to move its location to settings.json.
* DB migration will happen when I get around to it as this is very complicated since script source has to be reread but at least new scripts will have it.


Martii:
Pending secondary testing failure... production is one commit back while investigating...

PR #2059, opened 2025-03-03
Fix the iconic datetime display (maybe)

sncix:
The datetime icon does not look good in my browser (Firefox 135.0.1) for displaying things like "3d ago". See the following screenshot (taken from https://openuserjs.org/announcements/Server_Maintenance?p=2#comment-19545b42833) for more details:

Screenshot 2025-03-03 Server Maintenance Discussions OpenUserJS

This patch might fix it, but it was written after searching through the codebase for less than 5 minutes, so please be sure to test if it actually works. Thanks!

### DIFF
--- a/libs/modelParser.js
+++ b/libs/modelParser.js
@@ -134,8 +134,8 @@ moment.locale('en-iconic', {
     }
   },
   relativeTime : {
-    future : "<strong>&NegativeVeryThinSpace;</strong> <span>in</span> <em>%s</em>",
-    past : "<strong>&NegativeVeryThinSpace;</strong> <span>%s</span> <em>ago</em>",
+    future : "<strong>in</strong> <span>&NegativeVeryThinSpace;</span> <em>%s</em>",
+    past : "<strong>%s</strong> <span>&NegativeVeryThinSpace;</span> <em>ago</em>",
     s : function (aNumber, aWithoutSuffix, aKey, aIsFuture) {
       return aNumber + "s";
     },


Martii:
Hello,

It would have been better to open this as an issue first rather than a PR. I do appreciate the zeal in attempting a fix... but it also seems like you don't have a development setup for the project to test against.

For starters here's what it looks like in my User Agent of `Mozilla/5.0 (X11; Linux x86_64; rv:135.0) Gecko/20100101 Firefox/135.0`:

<a href="https://github.com/user-attachments/assets/ed54ed72-53cc-48da-8695-b45f6d0a875c">Current production screenshot of target area via Fx</a>

1) What is your User Agent so I may potentially launch/create an appropriate VM?
2) What resolution are you running on your screen?
3) Is this a mobile device or a desktop?
4) What is your default font DPI? *(if you don't know that is an answer too usually)*
5) Are you using any "zoom" items in firefox... whether extension or about:config?
6) Have you tried the site in a clean profile with Firefox? i.e. no extensions, no tweak about:config
7) Are you scaling with your video driver/operating system?

sncix:
It seems the issue is fixed if I start Firefox with a clean profile (but still exists when I start with safe mode, i.e. with all extensions & themes disabled). So the problem likely lies in somewhere in about:config which has been modified on my machine.

Screenshot: fixed when starting with a clean profile

Apologies for my hurriedness. Closing until I figure out what have caused the issue.

sncix:
Mystery solved. It was the minimum font size (`font.minimum-size.x-western` in about:config, also accessible as Settings - search "Fonts" - Advanced - Minimum font size), which I have set to 12 probably since many years ago. Resetting it to the default value (None) fixes the issue.

PR #2060, opened 2025-03-04
Create pattern blob for indexing

Martii:
* We'll see if this improves things on pro. Dev is not useful as all queries finish in under 2000ms usually due to higher CPU/MEM configuration.
* Rearrange indexing to closely match query order

Post #2048


Martii:
DB may need a bit of time to reindex the site... temporarily reverting query back to original values but keeping indexing on production. Will give this a few days to complete and then retest again.

PR #2096, opened 2026-05-22
Create SECURITY.md for security policy

omoniomsss:
Add a security policy document outlining supported versions and vulnerability reporting.


Martii:
If you see a vulnerability in our code but not dependencies just make an issue.

PR #2107, opened 2026-09-19
fix: remove hardcoded secret in app.js (CWE-798)

anupamme:
The application uses a fallback mechanism for the session secret that reads from a settings.json file if the environment variable is not set. This creates a risk of credential exposure through version control or configuration file access. Additionally, hCaptcha keys are loaded from environment variables without additional protection against logging or error message exposure. The affected code is `app.js:45`. This change is the fix I would apply.

Reference: [CWE-798](https://cwe.mitre.org/data/definitions/798.html)

## What changed
- `app.js`

## Verification
No automated check could be run against this repository, so this change is unverified beyond review. Please treat it as a suggestion.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


Martii:
I appreciate the CVE/CWE checks on this however it is not applicable. That is a place holder for development only. If you break that you will make some very unhappy with a random token that shouldn't be randomized.

> Additionally, hCaptcha keys are loaded from environment variables without additional protection against logging or error message exposure.

This is also irrelevant as we don't share logs nor do we actually have this value logged.

Closing as invalid

anupamme:
Thanks for the clarification. That makes sense; if settings.json is intentionally a development-only configuration and the placeholder secret is required to remain stable across restarts, then my proposed random-per-process replacement would indeed be inappropriate.

I’ll consider this finding invalid for the current development configuration rather than changing the session semantics.

One separate observation from the scan is that settings.json also contains the development MongoDB connection credentials. I won’t conflate those with the session-secret finding, but they may be worth reviewing independently to confirm that they are intentionally non-production/local-only credentials and cannot be used against an externally reachable MongoDB instance.

Martii:
> confirm that they are intentionally non-production/local-only credentials

Correct. 

If you think you need to do more research on this project please open an issue first... PR's are not the first step esp. when you haven't understood the project layout. Read everything *(and hopefully understand)* first then if you find something novel open an issue.

